PR #3, opened 2017-01-16
Return the disassembled code

badboy:
Instead of printing, return the array of instructions. This lets the user decide how and where to print it.


qmonnet:
Thanks!
But if we do that, we can as well delete the function, because the only thing it does is calling `to_insn_vec()` which is also public and printing its content (I'm realizing the checks on length at its beginning are redundant with those in `to_insn_vec()` and should be removed).

So with the current API, if the user wants to handle the vector of instructions they only have to call `to_insn_vec(&prog)` instead of `disassemble(&prog)` (see the example `to_json.rs`, this is exactly what happens). Do you think we should change that? The name of the functions maybe?

badboy:
Oh indeed, in that case we could just rename `to_insn_vec` to `disassemble` to make it obvious and printing can be handled in code using this API

qmonnet:
Would be fine for me. Just a note before we do that: I also added a [`to_insn_vec()`](https://github.com/qmonnet/rbpf/blob/master/src/ebpf.rs#L476) function to the `ebpf` module, and it does exactly the same as its counterpart in `disassembler` module, on a lower level (not grouping values for 16-bytes long `lddw`, not adding name and description). As they do practically the same I thought it was helpful to have the same name, but in different modules. I wrote it in case some users would like to print the 8-bytes insns one by one, or something like that.

So if we rename `disassembler::to_insn_vec()` into `disassembler::disassemble`, do you have a suggestion for renaming the one in `ebpf` module? Or for what to do about it?

badboy:
Good question. I have to re-read the code for that first.

PR #4, opened 2017-01-16
Use slices & references

badboy:
As promised here are a few Rust-y details to your code.
I ran [cargo-clippy](https://github.com/Manishearth/rust-clippy) against the code and fixed nearly all of the warnings.
A list of some of the changes:

1. Slice instead of Vec. [`ptr_arg`](https://github.com/Manishearth/rust-clippy/wiki#ptr_arg)
2. Don't re-borrow slices. [`needless_borrow`](https://github.com/Manishearth/rust-clippy/wiki#needless_borrow)
3. Useless use of vec. [`useless_vec`](https://github.com/Manishearth/rust-clippy/wiki#useless_vec)
4. …and a few more. [Full log of clippy against the master branch](https://gist.github.com/badboy/dee55394f010597413cd9ae806da0953)

I even changed `&mut Vec` to `&mut []` (i.e. a mutable vector to a mutable slice, as afaik it should not be possibleto resize the passed array, but only modify it).
Tests still run just fine.
If above assumption is not valid, I need to switch it back.

I'd be happy to answer questions regarding any of the changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -380,7 +380,7 @@ fn main() {
         None => panic!("Failed to look up .classifier section"),
     };
 
-    let ref prog = &text_scn.data;
+    let prog = &text_scn.data;
 
     // This is our data: a real packet, starting with Ethernet header
     let mut packet = vec![
@@ -409,7 +409,7 @@ fn main() {
     // We must provide the offsets at which the pointers to packet data start
     // and end must be stored: these are the offsets at which the program will
     // load the packet data from the metadata buffer.
-    let mut vm = rbpf::EbpfVmFixedMbuff::new(&prog, 0x40, 0x50);
+    let mut vm = rbpf::EbpfVmFixedMbuff::new(prog, 0x40, 0x50);
 
     // We register a helper function, that can be called by the program, into
     // the VM.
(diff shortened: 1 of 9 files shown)


qmonnet:
Thanks a lot, all of this looks great! I did not know about cargo-clippy, I'll start using it now I think.

I don't have time to review the changes in detail right now. I have to give some thought to the Vec to slice, we don't need to modify the size for now but I don't know if we are susceptible to do so in the future or not (maybe if we try to patch “context accesses” to be compatible with clang-compiled programs?)—this is why I kept vectors so far. I had not “seen” the other improvements, and on a first glance, it looks awesome! I'll try to read it more in detail soon.

badboy:
Wherever you were using immutable vectors, now using slices is strictly more general (not only Vectors can coerce to &[u8]), so it's safe to keep them there (plus using a vector through a reference is one addition indirection to the data).
Extending a Vector is only (safely) possible through it's methods anyway, so it could easily be changed back once it is necessary.

qmonnet:
Ok, thanks for those details. I'll probably merge the PR as is then, just allow me to review it at home tonight :).

Since the Vectors coerce, we do not need (but could, though) to change the unit test, right?

PR #6, opened 2017-01-20
[wip] API to build BPF programs

Alex-Diez:
Hi, I want to learn BPF too :smile:

I decided to start from instruction set. This PR was created just to share my idea how API of generating BPF instruction could look like.

I look forward to hearing your opinion about it.


instructions/src/lib.rs
```diff
@@ -0,0 +1,162 @@
+pub enum MemSize {
+    Double,
+    Half
+}
```
Alex-Diez: I will come to other two enum variants some day with TDD :wink:

instructions/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+            ins.push((self.offset & 0x00_FF) as u8);
+            ins.push(((self.offset & 0xFF_00) >> 8) as u8);
+            ins.push((self.immediate & 0x00_00_00_FF) as u8);
+            ins.push(((self.immediate & 0x00_00_FF_00) >> 8) as u8);
+            ins.push(((self.immediate & 0x00_FF_00_00) >> 16) as u8);
+            ins.push(((self.immediate & 0xFF_00_00_00) >> 24) as u8);
+        }
+        self.program
+    }
+}
```
Alex-Diez: I am not sure about that yet. I wanted to write a BPF program by calling chain of functions.
For instance:
```rust
program.bind_value().dst(0x00).value(0x00_00_11_22).push().bind_value().dst(0x01).value(0x11_22_00_00).push().add_registers().dst(0x00).src(0x01).push() ... /* and so forth */.exit()
```
Here I called function `push` referring _push on stack_, however I am not sure that it is a good name for it.

However, I have not considered to write it in this way:
```rust
program.add(Instruction::new(Class::Move).dst(0x00).src(0x01));
```

instructions/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+        self.instructions.as_slice()
+    }
+
+    pub fn bind_value(&'p mut self) -> Operation<'p> {
+        Operation::new(self, 0xb7)
+    }
+
+    pub fn bind_register(&'p mut self) -> Operation<'p> {
+        Operation::new(self, 0xbf)
+    }
```
Alex-Diez: I will change to `move`, it is more domain specific name

Alex-Diez:
> I take it you would eventually merge it in the crate, and not as a separate crate as in this PR, is this correct?
Yes, it is just faster to run `cargo test` on separated crate :smile:

> (keep in mind I am no Rust expert, I may be wrong)
Me either, I just write tests for each functionality that I want to implement and run `cargo test` as frequently as possible to make sure that compiler is not arguing and everything works as I want :wink:
 
I read through you comments ... Anyway I am developing in TDD so names and missed `MemSize`'s variants will be fixed eventually. Actually, by this PR I wanted to show you chain API and hear your opinion about it. I have rewritten one of you test to show how it looks like in real world. That's it.

Long names don't bother me a lot for now. Auto-complition does its job well.

qmonnet:
Yeah, the new version of the test is way more readable than the slice of hexa code… I'd be glad to merge your API once it's ready.

> Long names don't bother me a lot for now. Auto-complition does its job well.

Sure, but then there's no need to clutter the doc with long names if you gain nothing by making them long ;-). In the case of `ProgramCodeBuild`, that was just a suggestion, it's not that long and I'm fine with it if you really want to keep it.

Alex-Diez:
Writing tests for load instructions I noticed something interesting. 
Instructions 

| Op code | Mnemonic |
| ------- | -------- | 
| 0x18 | lddw dst, imm |
| 0x20 | ldabsw src, dst, imm |
| 0x28 | ldabsh src, dst, imm |
| 0x30 | ldabsb src, dst, imm |
| 0x38 | ldabsdw src, dst, imm |
| 0x40 | ldindw src, dst, imm |
| 0x48 | ldindh src, dst, imm |
| 0x50 | ldindb src, dst, imm |
| 0x58 | ldinddw src, dst, imm |

according LD/LDX/ST/STX opcode structure
```
msb      lsb
+---+--+---+
|mde|sz|cls|
+---+--+---+
```

have `000` class, but other loads have `001` class.

Instructions in the above table, are they some kind of special? Does their class have special name in terms of BPF?

qmonnet:
That's it, the difference is the source of the load operand, immediate or register. As you surely noticed already, it uses these definitions from `src/ebpf.rs`:

    pub const BPF_LD    : u8 = 0x00;
    pub const BPF_LDX   : u8 = 0x01;

(If I remember correctly the previous BPF version had a register called `X` for this kind of operations, that's probably where we get the name from.)

Alex-Diez:
Pushed second version of Rust API.

I have a couple of questions:
 * should I rewrite unit/integration tests with my API, if it is Ok?
 * should I merge ebpf::Insn structure and related functions with my API, if it is Ok?

instructions/src/lib.rs
```diff
@@ -1,162 +1,1532 @@
...
+    address_source: AddressSource,
     dst: u8,
     src: u8,
     offset: u16,
-    immediate: u32
+    imm: u32
+}
+
+impl Load {
+    pub fn new(mem_size: MemSize, addressing: Addressing, address_source: AddressSource) -> Self {
```
Alex-Diez: I am not sure about this `Addressing` and `AddressSource` enums, though. 😕

qmonnet:
Hey thanks for the hard work, that's impressive! Please can you give me a few days so that I can study and play with it? I don't have too much time right now to dive into the details.

Please do not start any rewriting of the tests coming from uBPF (`tests/upbf_*`) for now. We should also have an assembler soon, and may want to build these programs with this assembler so as to better follow the evolution of uBPF test suite. Once merged, your API could be used for all other unit tests—that's not so many tests for now, but I wish to extend the list, we have many uncovered error cases.

qmonnet:
OK so, sorry for the delay, here's some feedback.

I don't really know how to say it with tact, but the thing is, I preferred your first version with chaining the commands. I know I told you having a program as an attribute in the `Operation` seemed strange to me, but on a second thought, I just believe I told you crap. I first thought that we would have to add this program reference to the `ebpf::Insn` struct to merge it all, but I realize your `Operation` is something different. I can't find a name that would ideally describe it, maybe an `InsnBuilder` to match the `ProgramBuilder`. Well anyway, I think it could have an `ebpf::Insn` attribute in the end, instead of having again all insn fields, and it would look great.

Regarding v2, I tend to find it quite verbose, and too “low-level”. I realize you've made a great effort to reconstruct the operation codes with the different class / memory / etc. codes, but I don't believe the user is interested in this level of details, and should have so many steps to build an instruction (for example, using `Move` to perform an addition or multiplication or other ALUs is not exactly intuitive). Plus we will have an assembler to produce bytecode from uBPF-like assembly, which is higher level than the v2 of the API… In contrast, the `.load_mem()` function from v1 were easier to use I think.

The idea to have a different struct for each operation type is good, though. Maybe we could push this further and restrict the operand-related methods to fill only the operands that are relevant for the current operation, raising an error/panic!ing otherwise?

So again, sorry for changing my mind, I realize you've done a lot on this v2. What's your opinion on the matter, do you think one version is specifically better than the other?

instructions/src/lib.rs
qmonnet: Probably easier to decouple the different kinds of load into several operation type instead of using this `Adressing` everywhere? Maybe also for the `AddressSource` I prefer what you had in your v1?

instructions/src/lib.rs
```diff
@@ -1,162 +1,1532 @@
...
+
+impl<'a> From<&'a Addressing> for u8 {
+    fn from(addressing: &'a Addressing) -> u8 {
+        match *addressing {
+            Addressing::Undef => 0x00,
+            Addressing::Abs => 0x20,
+            Addressing::Ind => 0x40
+        }
+    }
+}
```
qmonnet: Not sure we have to use a match for extracting values from enums. Apparently you can give each item a _discriminant_ value, and then cast as an integer type, see <https://doc.rust-lang.org/reference.html#enumerations>.

instructions/src/lib.rs
```diff
@@ -1,162 +1,1532 @@
...
+
+pub struct Jump {
+    cmp: Comparison,
+    src_bit: Source,
+    dst: u8,
+    src: u8,
+    offset: u16,
+    imm: u32
+}
```
qmonnet: `imm` and `offset` are considered as signed. We'll have to fix it at some point if we use `ebpf::Insn`.

instructions/src/lib.rs
Alex-Diez: Thanks. Didn't know that.

Alex-Diez:
Thanks for the review.

First of all v2 was just a try :smile: as v1. The more Rust code I write the better Rust developer I become. That's one of a reason why I start contributing to this project :wink:

When I finished v2 I thougth that I would have to merge it with `ebpf::Insn` somehow. However, after #9 PR I think we need to develop `Instruction API` which will be the same for `ebpf`, `disassembler` and `asm_parser` modules and Rust BPF Builder API (this PR). What do you think about that?

qmonnet:
Hi @Alex-Diez, and sorry for being so slow to answer, I'm on a business trip and the WiFi is awful…

It might be nice to unify the instructions for API/assembler/disassembler, I haven't given a thought into that so far. But I'm not sure I see what you have in mind with this `Instruction` API, what would it look like and what would it bring exactly?

I'm being very cautious with the instructions in `ebpf` module, though. As I see it, this module should remain self-contained and only define stuff that is generic for eBPF. As such, it could ultimately be reused in other projects that need eBPF stuff but have nothing to do with userspace VM or (dis)assemblers. I don't know if this can have some impact on what you had in mind, I prefer to warn just in case :)

Alex-Diez:
Pushed last changes.

I left some todos.

1. I think `BpfCode` is better to call `RawBpfCode`. 
2. Add static functions (I am not sure about arguments types):
    -  `from_elf(path_elf: &str)`
    - `parse(asm: &str)`
3. Create `VerifiedBpfCode` struct, which will be created by calling `fn verify() -> Result<VerifiedBpfCode, VerificationError>`. 

For example:

```rust
let bpf_code = RawBpfCode::from_elf("/path/to/elf_file");
let verified_code = bpf_code.verify().unwrap();

let vm = rbpf::EbpfVmNoData::new(verified_code.assemble());
vm.prog_exec();
```

instructions/src/lib.rs
```diff
@@ -1,731 +1,1154 @@
...
+//        todo: impl verification
+//    }
+
     pub fn new() -> Self {
         BpfCode { instructions: vec![] }
     }
 
-    pub fn as_bytes(&self) -> &[u8] {
-        self.instructions.as_slice()
+    pub fn mov_add(&mut self, source: Source, arch: Arch) -> Move {
```
qmonnet: The `mov_` prefix, here and in the following functions, is misleading to new eBPF users. I understand you refer to the operation class that is used to form the operation code, but I don't think the user should have to care about that. Thoughts?

Alex-Diez:
First of all, I don't want to have two disassemble APIs. 😄 

I can remove `Disassemble` trait and merge into a separate module.

PS: working on this API I start thinking about this `RawBpfCode` and `VerifiedBpfCode` and let myself create needless functionality. And I understood the reason. It is because my code is OOP-like rather then procedural-like as in the other modules. 😄 
I start to imagine that we could write chaining code for BPF program in general. For instance:

``` rust
let bpf_vm = ... // create VM of any type

RawBpfCode::new().load(...).push().store(...).push()
// or RawBpfCode::parse(...)
// or RawBpfCode::from_elf(...)
.verify().map(|verified| verified.execute_with(bpf_vm)).err().map(...);
```

and `RawBpfCode::parse()` can actually use `asm_parser` functionality to be synched with it.

PS2: Maybe it is better to create an `Issue` and move this discussion there or should I just provide PR without `Disassemble` trait with current API?

qmonnet:
> I can remove Disassemble trait and merge into a separate module.

Sure, thanks!

> I start to imagine that we could write chaining code for BPF program in general.

Well, we most probably _could_. But is this something we _should_ do? This is a more OOP approach indeed, but does it make the lib easier to use? I mean, we already have the “procedural-like” approach that works well, why change it (real question, I'm open to listening to arguments)? Opening an issue about may be a good idea indeed. I am not sure anyone else will immediately take part in the discussion, but this is something we can leave open a while and see if other people express interest in it.

src/rust_api.rs
```diff
@@ -0,0 +1,1561 @@
...
+    instructions: Vec<u8>
+}
+
+impl BpfCode {
+
+    pub fn new() -> Self {
+        BpfCode { instructions: vec![] }
+    }
+
+    pub fn mov_add(&mut self, source: Source, arch: Arch) -> Move {
```
qmonnet: Again, I'm not sure the `mov_` prefix will help end users understanding what the function does.

src/rust_api.rs
```diff
@@ -0,0 +1,1561 @@
...
+                imm: 0x00_00_00_00
+            }
+        }
+    }
+
+    pub fn swap_bytes(&mut self, endian: Endian) -> SwapBytes {
+        SwapBytes {
+            bpf_code: self,
+            endian: endian,
+            insn: Insn {
```
qmonnet: I guess I should implement or derive the `Default` trait for `ebpf::Insn`…

src/rust_api.rs
```diff
@@ -0,0 +1,1561 @@
...
+pub enum Source {
+    /// immediate field will be used as a source
+    Imm = 0x00,
+    /// src register will be used as a source
+    Reg = 0x08
+}
+
+#[derive(Copy, Clone)]
+enum OpBits {
+    Add = 0x00,
```
qmonnet: Please use the values from `ebpf` module instead: `ebpf::BPF_ADD` and so on.

src/rust_api.rs
```diff
@@ -0,0 +1,1561 @@
...
+    Negate = 0x80,
+    Mod = 0x90,
+    BitXor = 0xa0,
+    NoOp = 0xb0,
+    SignRShift = 0xc0
+}
+
+#[derive(Copy, Clone)]
+pub enum Arch {
+    X64 = 0x07,
```
qmonnet: Same comment here.

src/rust_api.rs
```diff
@@ -0,0 +1,1561 @@
...
+
+    fn immediate(mut self, imm: i32) -> Self {
+        self.insn.imm = imm;
+        self
+    }
+}
+
+#[derive(Copy, Clone)]
+pub enum Endian {
+    Little = 0xd4,
```
qmonnet: And again.

src/rust_api.rs
```diff
@@ -0,0 +1,1561 @@
...
+
+    /// sets source register
+    fn src_reg(self, src: u8) -> Self;
+
+    /// sets offset bytes
+    fn offset_bytes(self, offset: i16) -> Self;
+
+    /// sets immediate value
+    fn immediate(self, imm: i32) -> Self;
+}
```
qmonnet: Most of these functions are exactly the same in all stucts implementing the trait. If I remember correctly, you can provide a “default” implementation in the trait declaration. With that we could avoid reimplementing identical functions a number of times, it's really a lot of lines for not so much gain…

src/rust_api.rs
```diff
@@ -0,0 +1,1561 @@
...
+
+    fn immediate(mut self, imm: i32) -> Self {
+        self.insn.imm = imm;
+        self
+    }
+}
+
+#[derive(Copy, Clone, PartialEq)]
+pub enum Cond {
+    Abs = 0x00,
```
qmonnet: …

src/rust_api.rs
```diff
@@ -0,0 +1,1561 @@
...
+    fn src_reg(self, src: u8) -> Self;
+
+    /// sets offset bytes
+    fn offset_bytes(self, offset: i16) -> Self;
+
+    /// sets immediate value
+    fn immediate(self, imm: i32) -> Self;
+}
+
+pub trait IntoBytes {
```
qmonnet: Yay, this name is way better than `assemble` here!

src/rust_api.rs
```diff
@@ -0,0 +1,1561 @@
...
+    pub fn mov(&mut self, source: Source, arch: Arch) -> Move {
+        self.mov_internal(source, arch, OpBits::NoOp)
+    }
+
+    pub fn mov_signed_right_shift(&mut self, source: Source, arch: Arch) -> Move {
+        self.mov_internal(source, arch, OpBits::SignRShift)
+    }
+
+    #[inline]
+    fn mov_internal(&mut self, source: Source, arch_bits: Arch, op_bits: OpBits) -> Move {
```
qmonnet: (No objection to keep `mov_` for this function, though)

tests/misc.rs
```diff
@@ -253,6 +253,8 @@ fn test_jit_block_port() {
     }
 }
 
+
+
```
qmonnet: You somehow left two additional empty lines here, please remove.

src/rust_api.rs
qmonnet: Also the names for the getters and setters are quite similar, and I find them confusing (`dst` to read the value, `dst_reg` to set it? And `offset_bytes` seems uselessly long). Could you find something else? Maybe adding `set_` prefix to the setters?

Alex-Diez: Yes, traits can have a default implementation. However, they don't have access to struct fields 😢

PS: AFAIR rust has some rfc for the abstract structs, but I could not find any.

src/rust_api.rs
```diff
@@ -0,0 +1,1561 @@
...
+    pub fn load_abs(&mut self, mem_size: MemSize) -> Load {
+        self.load_internal(mem_size, Addressing::Abs, 0x00)
+    }
+
+    pub fn load_ind(&mut self, mem_size: MemSize) -> Load {
+        self.load_internal(mem_size, Addressing::Ind, 0x00)
+    }
+
+    pub fn load_x(&mut self, mem_size: MemSize) -> Load {
+        self.load_internal(mem_size, Addressing::Undef, 0x61)
```
Alex-Diez: BPF_MEM = 0x60 ... but I can find the meaning of 0x01 ... only BPF_LDX 🤔

src/lib.rs
```diff
@@ -28,6 +28,7 @@ pub mod ebpf;
 pub mod helpers;
 mod jit;
 mod verifier;
+pub mod rust_api;
```
Alex-Diez: `code_build`, `insn_build` 🤔

src/rust_api.rs
qmonnet: Yes, that's `BPF_LDX` here I guess. In `ebpf` module you have for example

    pub const LD_B_REG   : u8 = BPF_LDX   | BPF_MEM | BPF_B;

(for “load byte from register” operation), or from Linux (linux/kernel/bpf/core.c):

    		[BPF_LDX | BPF_MEM | BPF_B] = &&LDX_MEM_B,

Alex-Diez: I had better make last changes on weekends ... Yesterday evening I saw `store_x` instead `load_x`. Hallucinations 😄

src/rust_api.rs
```diff
@@ -0,0 +1,1561 @@
...
+            }
+        }
+    }
+
+    pub fn store(&mut self, mem_size: MemSize) -> Store {
+        self.store_internal(mem_size, 0x00)
+    }
+
+    pub fn store_x(&mut self, mem_size: MemSize) -> Store {
+        self.store_internal(mem_size, 0x61)
```
Alex-Diez: Nope, I didn't have hallucinations yesterday 😄. Here, I didn't know what constant I should use. `BPF_MEM | BPF_STX` works.

src/rust_api.rs
qmonnet: I see, that's too bad… I'm sure there is something we can do to refector this. But then it can be done later, when we think about something clean I guess.

src/rust_api.rs
qmonnet: Well, maybe you simply should not have 0x61 there? I think it should be `BPF_STX | BPF_MEM`, so 0x63 maybe? I haven't checked whether the tests pass if we change this value.

src/lib.rs
qmonnet: Hmm I don't know. I don't like `code_build` much, we're not building “code”. I would accept `insn_build` I think. We can keep searching for a name until we merge. Should ideally be short, for people that do not include `<module>::*` but will prefer to call `<module>::<function>()` a lot.

Alex-Diez:
Forget to add changed file :bowtie:

src/insn_builder.rs
```diff
@@ -0,0 +1,1389 @@
...
+        buffer.push(self.get_src() << 4 | self.get_dst());
+        buffer.push((self.get_off() & 0x00_ff) as u8);
+        buffer.push(((self.get_off() & 0xff_00) >> 8) as u8);
+        buffer.push((self.get_imm() & 0x00_00_00_ff) as u8);
+        buffer.push(((self.get_imm() & 0x00_00_ff_00) >> 8) as u8);
+        buffer.push(((self.get_imm() & 0x00_ff_00_00) >> 16) as u8);
+        buffer.push(((self.get_imm() & 0xff_00_00_00) >> 24) as u8);
+        buffer
+    }
+}
```
qmonnet: Two warnings are raised by this function, regarding the int values that are used for the masks on lines 79 and 83. In fact, all those masks are useless, and you can do as follows:

```
    fn into_bytes(self) -> Self::Bytes {
        let mut buffer = Vec::with_capacity(8);
        buffer.push( self.opt_code_byte());
        buffer.push( self.get_src() << 4 | self.get_dst());
        buffer.push( self.get_off()        as u8);
        buffer.push((self.get_off() >> 8)  as u8);
        buffer.push( self.get_imm()        as u8);
        buffer.push((self.get_imm() >> 8)  as u8);
        buffer.push((self.get_imm() >> 16) as u8);
        buffer.push((self.get_imm() >> 24) as u8);
        buffer
    }
```

Casting to `u8` after the division automatically trims the unwanted part on the “left”.

(Of course you don't have to keep the same formating, aligning stuff is just my personal preference in such case).

src/insn_builder.rs
```diff
@@ -0,0 +1,1389 @@
+// Copyright 2017 <alex.dukhno@icloud.com>
```
qmonnet: Maybe with your name in addition to the email?

    // Copyright 2017 Alex Dukhno <alex.dukhno@icloud.com>

qmonnet:
Ok, warnings are gone, I think we're good this time! :). The commit history on this PR is a bit chaotic though, so I intend to squash commits. Is is ok for you, or do you prefer to split your changes yourself into a new set of (logical) commits?

Alex-Diez:
Just squash it

qmonnet:
Done!

qmonnet:
100th commit, by the way

Alex-Diez:
Anniversary 😄

PR #9, opened 2017-02-06
add assembly parser

rlane:
Uses the combine crate to parse ubpf-style assembly to Instruction
structs. A future assembler module will consume these, check that
the instruction and operands are valid, and output ebpf::Insn.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,10 +21,10 @@ use std::collections::HashMap;
 extern crate libc;
 extern crate combine;
 
+pub mod asm_parser;
 pub mod disassembler;
 pub mod ebpf;
 pub mod helpers;
-pub mod asm_parser;
 mod jit;
 mod verifier;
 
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -19,10 +19,12 @@ use std::u32;
 use std::collections::HashMap;
 
 extern crate libc;
+extern crate combine;
 
 pub mod disassembler;
 pub mod ebpf;
 pub mod helpers;
+pub mod asm_parser;
```
qmonnet: Ah also, can we keep includes in alphabetical order (in each one of the crate/public module/private module section)? Also, for dependencies in the `Cargo.toml` file?

rlane:
Thanks for the pointer to rust-clippy. Pushed a few commits.

qmonnet:
Thanks a lot! Merging.

PR #10, opened 2017-02-12
eBPF assembler

### LINKED ISSUE #7: Assembler
**Please, do not start working on this**, someone already asked me by email to start the work on the assembler.

New feature: an assembler similar to the one of uBPF, to build eBPF programs with a syntax similar to:

```
mov32 r0, 0
mov32 r1, 2
add32 r0, 1
add32 r0, r1
exit
```

This would make rbpf natively compatible with uBPF test cases (most functional unit tests have been translated into bytecode already, but it would make it more readable to have them in human-readable form, and it would help be compatible when uBPF receives new test cases.

Improvement suggestion: use `add64` instead of `add` for 64-bits arithmetic operations (but also accept `add` only, to remain compatible).

rlane:
Supports the same syntax as ubpf, plus "64" suffixes for ALU instructions. Doesn't yet support exotic instructions like TAIL_CALL or ST_W_XADD.

A couple of tests are changed to use the assembler instead of binary as a proof of concept. We can do the rest in another pull request.

Fixes #7.


src/assembler.rs
```diff
@@ -0,0 +1,225 @@
...
+        return Err(format!("Invalid immediate {}", imm));
+    }
+    Ok(Insn {
+        opc: opc,
+        dst: dst as u8,
+        src: src as u8,
+        off: off as i16,
+        imm: imm as i32,
+    })
+}
```
qmonnet: This whole function is pretty close to a simplified version of the program building API @Alex-Diez has been working on in PR #6. Once we have the API, we could use it here instead to avoid re-defining ways to build instructions.

src/assembler.rs
```diff
@@ -0,0 +1,225 @@
...
+        }
+        (JumpConditional, Register(dst), Integer(imm), Integer(off)) => {
+            insn(opc | ebpf::BPF_K, dst, 0, off, imm)
+        }
+        (Call, Integer(imm), Nil, Nil) => insn(opc, 0, 0, 0, imm),
+        (Endian(size), Register(dst), Nil, Nil) => insn(opc, dst, 0, 0, size),
+        (LoadImm, Register(dst), Integer(imm), Nil) => insn(opc, dst, 0, 0, (imm << 32) >> 32),
+        _ => Err(format!("Unexpected operands: {:?}", operands)),
+    }
+}
```
qmonnet: Probably our major difference in formatting style: I'd definitely align the patterns on something like this :) I find it more readable, but it's not really a good habit since later modifications of a line can force to realign everything (so I'm not asking you to align, that's just for the record).

tests/assembler.rs
```diff
@@ -0,0 +1,428 @@
...
+fn test_error_operands_out_of_range() {
+    assert_eq!(asm("add r16, r2"),
+               Err("Failed to encode add: Invalid destination register 16".to_string()));
+    assert_eq!(asm("add r1, r16"),
+               Err("Failed to encode add: Invalid source register 16".to_string()));
+    assert_eq!(asm("ja -32769"),
+               Err("Failed to encode ja: Invalid offset -32769".to_string()));
+    assert_eq!(asm("ja 32768"),
+               Err("Failed to encode ja: Invalid offset 32768".to_string()));
+    assert_eq!(asm("add r1, 4294967296"),
```
qmonnet: Well then, since you use the same value again I suppose this is not a mistake… Let me guess, user can consider the immediate value as a signed or as an unsigned? That would make sense indeed. Is there a reason why we should not allow the same for the offset in that case?

src/assembler.rs
```diff
@@ -0,0 +1,225 @@
...
+        return Err(format!("Invalid destination register {}", dst));
+    }
+    if dst < 0 || src >= 16 {
+        return Err(format!("Invalid source register {}", src));
+    }
+    if off < -32768 || off >= 32768 {
+        return Err(format!("Invalid offset {}", off));
+    }
+    if imm < -2147483648 || imm >= 4294967296 {
+        return Err(format!("Invalid immediate {}", imm));
```
rlane: I'd been assuming we need to support 32-bit unsigned integer constants. Thinking more about it, constants with the high bit set are going to be sign-extended by most instructions and allowing positive-looking constants that are interpreted by the VM as negative would be confusing. I pushed a commit fixing this.

I initially used std:::i32::MIN etc, but found that they were sign-extended when casting to i64. Rather than doing a cast plus a mask, I thought it would be more readable to write out the value.

src/assembler.rs
rlane: Yeah, I'd normally align it but I've been using rustfmt which would delete the extra spaces.

rlane:
Sure, I can add '#' syntax for comments in a future pull request.

I also plan to send a pull request that changes (automated) the ubpf tests to use the assembler.

src/assembler.rs
qmonnet: … because we have to parse numbers as `i64`… Ok, I see. Well, fine with me, then. Also, forbidding unsigned immediate sounds like a good idea, and we can still revert this later if it appears that it is really needed. Let's merge all this!

qmonnet:
Thanks a lot! Merging now

PR #11, opened 2017-02-20
Use assembler in testcases

rlane:
The conversion is mostly automatic. It turns out the ubpf testcases do use positive immediates with the high bit set. I think I'll change the testcases in the ubpf repo since the rbpf behavior makes more sense.


qmonnet:
Looks perfect, thanks a lot! This will make it easier to stay in sync with uBPF, or to simply read the programs.

PR #15, opened 2017-04-21
Correct disasm

badboy:
Debugging my own BPF programs I noticed a wrong output in the disassembler.
This fixes it, see the first two commits.

I added tests for the disassembler output by roundtripping through the assembler (so it's easier to write). I more or less took all assembler tests, only removed the error-case ones and some which used shortened instructions (`add` instead of `add64`). This testing is not perfect, but should catch a mismatch here.

### DIFF
--- a/src/disassembler.rs
+++ b/src/disassembler.rs
@@ -22,11 +22,11 @@ fn alu_reg_str(name: &str, insn: &ebpf::Insn) -> String {
 
 #[inline]
 fn byteswap_str(name: &str, insn: &ebpf::Insn) -> String {
-    match insn.off {
+    match insn.imm {
         16 | 32 | 64 => {},
         _ => println!("[Disassembler] Warning: Invalid offset value for {} insn", name)
     }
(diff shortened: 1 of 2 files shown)


qmonnet:
Good catch, thanks! The `imm` <-> `off` confusion is a stupid mistake of mine, and it seems I missed the difference for load and store at the time I wrote the disassembler. Thanks!

I'll review the tests this weekend. I saw Rich's copyright in the file boilerplate, did you reuse tests from uBPF?

badboy:
Oops, I actually just copied over `test/assembler.rs`. I can remove that statement

qmonnet:
Yeah, well, probably better add your own copyright instead :-)

badboy:
Will do

badboy:
Changed the copyright

qmonnet:
Tests are fine as well, thanks for that. Thanks for the fix. Thanks for the PR. Thanks for everything! Merging now.

PR #16, opened 2017-05-02
Remove no-longer-existing lint

badboy:
This was recently removed from Rust. With this patch there is no "unknown lint" warning on Rust 1.17 anymore.
Older versions should still work, though maybe with a warning about `single_match`

I'm not sure what `single_match` would have complained about before.


qmonnet:
If I remember correctly, the `single_match` was triggered by the `match` blocs with a single pattern inside that could be matched (plus the default `_ => {}`, that would not do much), and it could probably have been replaced by `if`s. In this case I wanted to keep the `match` bloc to have the same algorithm structure for `OperandSize::S16` as for the other sizes just below.

Anyway, a lint warning for old rustc versions seems totally acceptable.

Thanks a lot for the fix! Made me upgrade my Rust setup by the way.

Issue #19, opened 2017-09-04
macOS build fails due to clock_gettime usage


waywardmonkeys:
macOS doesn't have `clock_gettime` or `CLOCK_MONOTONIC`, so the `bpf_time_getns` helper function fails to compile on macOS.

qmonnet:
Thanks for the report. I don't have a machine running macOS at hand so I've never tried compiling rbpf on this OS, would you know what function we could use instead?

waywardmonkeys:
I was just looking at preparing a patch that uses `std::time::{Instant,Duration}` (https://doc.rust-lang.org/std/time/struct.Instant.html) instead ...

waywardmonkeys:
Well, that won't work... but I'll submit a pull request that fixes on macOS within a day.

qmonnet:
No hurry here, thanks a lot!

badboy:
While at it we could add MacOS as a target on Travis to make sure it continues to work

qmonnet:
Yeah, that's a good idea. We can do that once this bug is fixed and we have something that compiles.

While we're at it, I've not tried on Windows, either. I don't know if it compiles (the `clock_gettime` probably fails as well, but what else?). Has anyone here tried to work with Rust on Windows?

badboy:
[AppVeyor](https://www.appveyor.com/) is free and easy to setup as CI as well to run the build. The jitting might be incompatible, but the rest should probably work.

waywardmonkeys:
My change to use the `time` crate will let that code at least work on Windows. I can't speak to any of the rest of it.

badboy:
I can take care of an AppVeyor config today

qmonnet:
@waywardmonkeys: Your patch works fine on Linux, thanks a lot! There is just a sentence that we should remove from the comment of the function, see review in the PR.

@badboy: Thanks a lot, I'd be glad to let you have a look. Tell me if you need some specific rights on the repo to make this work.

I really appreciate your help and motivation! :-)

waywardmonkeys:
@qmonnet I'll fix up the PR shortly. Dinner time here!

qmonnet:
No problem. _Bon appétit_!

PR #20, opened 2017-09-04
Use time crate for nanosecond times

waywardmonkeys:
This uses the `time` crate, which is marked as deprecated, but it is still maintained. They suggest using `chrono` instead, but that's far larger and what's needed here is so small that it isn't reasonable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ rust:
   - stable
   - beta
   - nightly
+os:
+  - linux
+  - osx
 matrix:
   allow_failures:
     - rust: nightly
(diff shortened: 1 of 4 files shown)


waywardmonkeys:
This also adds macOS to Travis as suggested in #19.

badboy:
LGTM

waywardmonkeys:
Force pushed with an update to the comment. Is that what you want?

badboy:
LGTM=Looks good to me.

badboy:
(Ah, only saw the comment in the other issue now)

qmonnet:
Perfect, I was only waiting for Travis builds to finish. Merging now!

PR #22, opened 2017-09-04
Lint and warning fixes

waywardmonkeys:
This gets closer to a clean build and clean clippy.


badboy:
LGTM, should we add a Clippy run to the Travis CI config?

qmonnet:
I don't have the time to test this now, but it looks fine. Merging. Thanks!

@badboy Seems today is the day we reinforce the CI tests, so why not go for a Clippy run as well :-)

PR #25, opened 2017-09-05
Remove redundant checks from disassemble().

waywardmonkeys:
These checks are already performed by to_insn_vec().


qmonnet:
Correct, thanks!

PR #26, opened 2018-03-24
Use [package] in Cargo.toml

waywardmonkeys:
This is what various tools expect, like `cargo outdated`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,4 @@
-[project]
+[package]
 
 # Project metadata
 name = "rbpf"


badboy:
Interestingly enough: `cargo` itself does accept both variants.

qmonnet:
What's the difference between the two keywords?

badboy:
Sorry for merging it that quickly.
[package](https://doc.rust-lang.org/cargo/reference/manifest.html) is the one that is mainly used by cargo and is also the one a `cargo new` will generate.

I suspect `[project]` is there for legacy reasons (seems there are test cases on the cargo repo for it). I right now don't have time to track it down further.

qmonnet:
Ok thanks. No problem for the merge, as long as you know what you're doing :)

badboy:
"Know what I'm doing" based on "it still compiles and is what cargo does", so yeah :)

qmonnet:
Ah ah, at least it still compiles—could be worse ;) Thank you both!

PR #27, opened 2018-09-06
fix clippy error: cast to more strict alignment, warning: not consumed copy

jackcmay:
Interested in using rBPF for a project I'm working on so fixed a couple of clippy errors/warnings.  If this project is still active and these changes are approved would like to fix more

- pulled byteorder dependency back in to avoid unsafe code and to fix clippy casting errors
- removed copy of operand that was not consumed

### DIFF
--- a/src/ebpf.rs
+++ b/src/ebpf.rs
@@ -18,6 +18,9 @@
 //! <https://www.kernel.org/doc/Documentation/networking/filter.txt>, or for a shorter version of
 //! the list of the operation codes: <https://github.com/iovisor/bpf-docs/blob/master/eBPF.md>
 
+extern crate byteorder;
+use self::byteorder::{ByteOrder, LittleEndian};
+
 /// Maximum number of instructions in an eBPF program.
 pub const PROG_MAX_INSNS: usize = 4096;
 /// Size of an eBPF instructions, in bytes.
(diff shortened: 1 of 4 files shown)


badboy:
/o\ I did? Will review it now. Sorry for the delay.

src/ebpf.rs
```diff
@@ -18,6 +18,9 @@
 //! <https://www.kernel.org/doc/Documentation/networking/filter.txt>, or for a shorter version of
 //! the list of the operation codes: <https://github.com/iovisor/bpf-docs/blob/master/eBPF.md>
 
+extern crate byteorder;
```
badboy: This should not be needed, it's imported on the root level already.

src/ebpf.rs
```diff
@@ -18,6 +18,9 @@
...
+use self::byteorder::{ByteOrder, LittleEndian};
```
badboy: You can drop the `self::` here then.

src/jit.rs
```diff
@@ -297,7 +297,7 @@ fn emit_load_imm(jit: &mut JitMemory, dst: u8, imm: i64) {
 
 // Store register src to [dst + offset]
 #[inline]
-fn emit_store(jit: &mut JitMemory, size: OperandSize, src: u8, dst: u8, offset: i32) {
```
badboy: We should simply make `OperandSize` copy in `src/jit.rs`

src/lib.rs
```diff
@@ -22,6 +22,9 @@
 use std::u32;
 use std::collections::HashMap;
 
+extern crate byteorder;
+use self::byteorder::{ByteOrder, LittleEndian};
```
badboy: Keep the `extern crates` together and move the `use byteorder...` line below that then. The `self::` should not be necessary anymore.

jackcmay: Would you prefer to keep all the `use`'s together and move the two existing`use std::` lines below the group of `extern`'s

badboy: That's my preferable style, yes.

1. extern
2. use std
3. use external stuff

Issue #30, opened 2018-10-04
Gracefully fail when possible, panic only if necessary


jackcmay:
Currently rbpf panics for any error.  Instead, where possible fail gracefully and return an error to the user.

This also has implications for the verification function.  Its current signature returns a bool but rbpf ignores it and expects the verification function to panic.  Part of addressing this issue is to respect the return value and potentially pass a more specific failure code up to the user.

qmonnet:
I merged #31, which addresses this point for the default verifier. There are more parts that could use proper error handling, let me know if you intend to work on that before we relase a crate (I will probably not do it myself at the moment), or if this should be left for later.

qmonnet:
This one was addressed with #37 right?

PR #31, opened 2018-10-06
Gracefully handle program rejection from verifier

qmonnet:
Submitting as a PR, because feedback from people better than myself at Rust would be welcome before I push :).

These two commits make the verifier return a `Result<(), String>` instead of panicking when a program does not pass verification tests. This is in order to avoid exiting the (Rust) program automatically when a (eBPF) program is rejected.

First commit wraps the call to `panic!()` for program rejection in a single function, which also adds the `"[Verifier] Error: "` prefix to the messages. Then the second commit changes this `panic!()` into an error that can be propagated upwards, and updates all tests with `unwrap()` each time this is fit. All tests pass (yeah!).

To do: This changes the basic API, the README file and the example invocation it contains will really need an update after we merge this. 

Addresses, at least in parts, issue #30.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,25 +132,26 @@ All these structs implement the same public functions:
 
 ```rust
 // called with EbpfVmMbuff:: prefix
-pub fn new(prog: &'a [u8]) -> EbpfVmMbuff<'a>
+pub fn new(prog: &'a [u8]) -> Result<EbpfVmMbuff<'a>, Error>
 
 // called with EbpfVmFixedMbuff:: prefix
 pub fn new(prog: &'a [u8],
            data_offset: usize,
-           data_end_offset: usize) -> EbpfVmFixedMbuff<'a>
+           data_end_offset: usize) -> Result<EbpfVmFixedMbuff<'a>, Error>
 
 // called with EbpfVmRaw:: prefix
-pub fn new(prog: &'a [u8]) -> EbpfVmRaw<'a>
+pub fn new(prog: &'a [u8]) -> Result<EbpfVmRaw<'a>, Error>
 
 // called with EbpfVmNoData:: prefix
-pub fn new(prog: &'a [u8]) -> EbpfVmNoData<'a>
+pub fn new(prog: &'a [u8]) -> Result<EbpfVmNoData<'a>, Error>
 ```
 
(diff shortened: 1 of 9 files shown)


jackcmay:
@qmonnet If this change requires a change to the package major version would you also like to get in the replaceable verifier changes as well since that also changes the api?

badboy:
This crate is at version 0.0.3, so... I guess we shouldn't worry too much about version numbers yet :D

jackcmay:
We would like to get an updated crate released by the end of the month that includes both these changes and the replacable verifier.  Is that possible?

qmonnet:
Hi, I have nothing against updating the crate (as long as we update the README before that, but that should not be difficult). I haven't done it in some time because I did not think anyone would use it, but if it helps I'll be glad to tag and push a new version.

I'm not sure I understood your remark regarding the major version number, do you have a reason why you would like to change it?

As for the API, I think I wrote somewhere that things were still susceptible to break. Which means I don't have a problem at the moment with changing the API - for returning an error for example instead of panicking, I'm sure that's actually a good thing. What I said regarding the custom verifier is not that I don't want to change things, but that I am reluctant to change the `new()` functions, because it should remain simple to use and I don't believe many users will try to plug a custom verifier so I'd as well avoid having to systematically have to pass a `None` parameter to `new()` if I can. On the same basis that I do not want to pass a list of helpers to register at the creation of the VM, for example, but I prefer to register them later, if that makes sense?

@badboy Thanks! I'll try to see if I can make it work with real errors before pushing.

qmonnet:
So I added a commit to change strings into error, but feedback would be welcome again. I'm unsure this is the way I should do it, putting formatted description such as eBPF instruction number in the error description might not be the best thing to do.

(Once it's ready I'll squash the commit with the previous one before merging)

qmonnet:
So I squashed my commits and pushed to master. I just forgot to update this PR and to push from here. So that's commits
* f6aae40d6918b09f531f82a44615546597b8217d
* fc834c71fb0344c69e6ff45b9c27a2174bae4cc5
* 541c88358361814c074affa098466b09cbc00376
Closing the PR now since the code is merged.

Issue #37, opened 2018-10-16
prog_exec has no way to report a vm error


jackcmay:
Callers may not want to panic if there is a problem running the program and instead may want to exit gracefully and report an error.

Consider changing:
```rust
pub fn prog_exec(&self, mem: &[u8], mbuff: &[u8]) -> u64
```

To:
```rust
pub fn prog_exec(&self, mem: &[u8], mbuff: &[u8]) -> Result<u64, Error>
```

jackcmay:
Can we get this in the same release as the other public interface changes?  Doing so would reduce the change of breaking backward compatibility in the future.

qmonnet:
Sure. I have no constraints regarding the release on my side anyway, so we only tag once you feel everything you need in the release is covered :).

For the record this is the logical continuation of #30.

jackcmay:
FYI, I'm working on this issue

jackcmay:
@qmonnet 

What do you think about returning a `String` instead of `Error`?  Doesn't look like Error is adding anything in this case.  Would also like to convert new, set_verifier, etc... to `String` 

```rust
-> Result<u64, String>
```
instead of
```rust
-> Result<u64, Error>
```

qmonnet:
Well I did the opposite, my first PR would use a `String` for the return values and I converted that to `Error`s based on suggestions I got :). What's the problem with `Error`s? This is something you can propagate and catch more easily than a string I would think?

jackcmay:
I see propagation and catch being handled by normal `Result` handling.  I guess I don't see the advantage of using `Error` in this case since only ErrorKind::Other is used and just used as a wrapper for the string without adding any additional information.  If there was a need to return other types of `ErrorKind` I could see the benefit but otherwise, I must be missing something :-)

jackcmay:
Maybe a better approach would be a module-specific error enum:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=090c45f8c4f4ec5c373a5f2cafdf3817

jackcmay:
@qmonnet I would like to get this issue resolved early this week so we can release.  I can submit my PR with changes to return Result<(), Error> that I have ready or I can rework to include my suggestion to return a string or custom Error enum.  Please let me know, thanks!

qmonnet:
Sorry for the late reply.

[...]

PR #38, opened 2018-10-23
Exit gracefully on program exit

jackcmay:


src/lib.rs
```diff
@@ -230,10 +228,11 @@ impl<'a> EbpfVmMbuff<'a> {
     /// // Register a helper.
     /// // On running the program this helper will print the content of registers r3, r4 and r5 to
     /// // standard output.
-    /// vm.register_helper(6, helpers::bpf_trace_printf);
+    /// vm.register_helper(6, helpers::bpf_trace_printf).unwrap();
     /// ```
-    pub fn register_helper(&mut self, key: u32, function: fn (u64, u64, u64, u64, u64) -> u64) {
+    pub fn register_helper(&mut self, key: u32, function: fn (u64, u64, u64, u64, u64) -> u64) -> Result<(), Error> {
```
qmonnet: Should we change `register_helper()`? Currently there's no way it can return an error, so maybe returning nothing is fine?

README.md
```diff
@@ -615,8 +617,8 @@ not trivial, and we cannot “copy” it since it is under GPL license.
 ### What about safety then?
 
 Rust has a strong emphasize on safety. Yet to have the eBPF VM work, some
-“unsafe” blocks of code are used. The VM, taken as an eBPF interpreter, can
-`panic!()` but should not crash. Please file an issue otherwise.
+“unsafe” blocks of code are used. The VM, taken as an eBPF interpreter can
```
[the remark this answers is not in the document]
qmonnet: Also could you please use backquotes around “`unsafe`”?

jackcmay:
I figured might as well enhance set_helper() along with the others in case there is a reason to return an error in the future.

src/lib.rs
```diff
@@ -50,6 +50,9 @@ mod verifier;
 ///   - Unknown eBPF helper index.
 pub type Verifier = fn(prog: &[u8]) -> Result<(), Error>;
 
+/// eBPF Jit-compiled program.
+pub type JitCompiled = unsafe fn(*mut u8, usize, *mut u8, usize, usize, usize) -> u64;
```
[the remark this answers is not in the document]
qmonnet: Not necessarily. But it's true that would be more consistent. Hmm... If you haven't done the change already, maybe use simply `JitProgram` instead of `JitCompiled`? Sounds more like a noun and less like an attribute. And then keep `Verifier` untouched? What do you think?

src/lib.rs
qmonnet: Ok, fine by me then.

qmonnet:
Ok, I buy it for `register_helper()`. So the PR looks good! With the nits addressed (I commented on the `JitCompiled` renaming), feel free to merge it.

README.md
```diff
@@ -565,7 +565,7 @@ Other than the language, obviously? Well, there are some differences:
 
 * When an error occur while a program is run by uBPF, the function running the
   program silently returns the maximum value as an error code, while rbpf
-  returns a String describing the error.
+  Rust type `Error`.
```
qmonnet: “Rust type“? Missing a verb?... :)

jackcmay:
Fixes #37

PR #39, opened 2018-10-24
Doc nits

jackcmay:
Ran README.md through a grammer checker, let me know if you agree with the changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -635,10 +635,10 @@ on your own.
 * The JIT compiler produces an unsafe program: memory access are not tested at
   runtime (yet). Use with caution.
 
-* If a division by 0 is attempted by a JIT-ted program, the program
-  returns `0xffffffffffffffff` and exits cleanly (no `panic!()`). This is
-  because the author has not found how to make `panic!()` work from the
-  generated assembly so far.
+* Contrary to the interpreter, if a division by 0 is attempted, the JIT program
+  returns `0xffffffffffffffff` and exits cleanly (no `Error` returned). This is
+  because the author has not found how to return something more explicit
+  from the generated assembly so far.
 
 * A very little number of eBPF instructions have not been implemented yet. This
   should not be a problem for the majority of eBPF programs.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -537,7 +537,7 @@ get more information and examples on how to use it.
 
 This is the author's first try at writing Rust code. He learned a lot in the
 process, but there remains a feeling that this crate has a kind of C-ish style
-in some places instead of the Rusty look the author would like it to have. So
+in some places instead of the Rusty look, the author would like it to have. So
```
qmonnet: Nack for this one, it's “the author would like it to have the Rusty look”, adding a comma does not sound right.

README.md
```diff
@@ -160,7 +160,7 @@ kernel). Users are also able to replace it with a custom verifier.
 For `struct EbpfVmFixedMbuff`, two additional arguments must be passed to the
 constructor: `data_offset` and `data_end_offset`. They are the offset (byte
 number) at which the pointers to the beginning and to the end, respectively, of
-the memory area of packet data are to be stored in the internal metadata buffer
+the memory area of the packet data, are to be stored in the internal metadata buffer
```
qmonnet: Nack for this one. it's “the offset at which the pointers (to the start and end (of memory area (of packet data))) are to be stored”. Adding a comma does not sound right?

README.md
```diff
@@ -635,7 +635,7 @@ on your own.
 * The JIT compiler produces an unsafe program: memory access are not tested at
   runtime (yet). Use with caution.
 
-* Contrary to the interpreter, if a division by 0 is attempted, the JIT program
+* If a division by 0 is attempted by a JIT-ted program, the program
```
qmonnet: The remark was more about the `panic` than the “contrary” :). Sorry, I didn't make that clear. Something like

```
Contrary to the interpreter, if a division by 0 is attempted, the JIT program
returns `0xffffffffffffffff` and exits cleanly (no `Error` returned). This is
because the author has not found how to return something more explicit
from the generated assembly so far.
```

Thanks for the commas.

PR #55, opened 2021-09-07
fix how the disassembler handles offsets

ttlajus:
As explained in the commit message of 038ddf7:

> https://doc.rust-lang.org/std/fmt/trait.LowerHex.html says "For primitive signed integers (i8 to i128, and isize), negative values are formatted as the two’s complement representation."
> 
> As such, we can't rely on Rust's "+" formatting flag when writing instruction offsets.


qmonnet (approved):
Sorry, I missed the notification for this PR and I'm only seeing it now.

The changes look all good, thanks a lot for this work!

PR #71, opened 2023-03-09
README: fix example type err

60ke:


60ke:
should not

qmonnet (changes requested):
Hi, and thanks!

Can you drop the unrelated changes (asterisks are fine for unordered lists in Markdown, no need to replace them with dashes) please?

How did you find these errors? I think the samples in the README.md used to be run as part of `cargo test`, but apparently this is not longer the case. Would be interesting to explain in the commit description, too.

Last thing: Over the last commits I've enforced the use of the [Developer Certificate of Origin](https://wiki.linuxfoundation.org/dco) for new contributions. I know this is not yet documented in the repository, but could you sign off your commit, please?

PR #76, opened 2023-05-10
deps: update time crate version.

ipuustin:
The time crate version used in rbpf [has a security issue](https://rustsec.org/advisories/RUSTSEC-2020-0071) (though the use of it in this library is secure). Update the time crate version to make the audits of the crates which depend on rbpf succeed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ include = [
 
 combine = "2.5"
 libc = "0.2"
-time = "0.1"
+time = "0.2"
 byteorder = "1.2"
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


qmonnet:
CI was failing because of #77 with recent versions of the toolchain, so I pushed a fix on `master`, and went ahead and rebased/force-pushed your branch on top of that.

qmonnet (approved):
Thanks! I had the same changes locally, but realising that the library is not affected, I never pushed them. If it helps with audits, then no objection.

I suppose you'd need a new version of the crate after that?

ipuustin:
Thanks! A new version would be nice at some point, but there is no real hurry about that. :-) I was examining libcontainer dependencies.

qmonnet:
https://crates.io/crates/rbpf/0.2.0 freshly released.

Issue #77, opened 2023-05-10
Tests panick with `misaligned pointer dereference: address must be a multiple of ... but is ...`


qmonnet:
Tests consistently fail with recent versions of the toolchain, with messages such as:

```
thread 'test_jit_mbuff' panicked at 'misaligned pointer dereference: address must be a multiple of 0x4 but is 0x7f8038001011', src/jit.rs:100:5
```

<details><summary>Full error message</summary>

```
     Running tests/misc.rs (target/debug/deps/misc-6f56f75d4e6d43c9)

running 20 tests
thread 'test_jit_mbuff' panicked at 'misaligned pointer dereference: address must be a multiple of 0x4 but is 0x7f8038001011', src/jit.rs:100:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'test_jit_mbuff' panicked at 'panic in a function that cannot unwind', library/core/src/panicking.rs:126:5
stack backtrace:
   0:     0x562d63bf4351 - std::backtrace_rs::backtrace::libunwind::trace::h28494931c73179b2
                               at /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x562d63bf4351 - std::backtrace_rs::backtrace::trace_unsynchronized::h9032c52edccf7bd1
                               at /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x562d63bf4351 - std::sys_common::backtrace::_print_fmt::hd90562e967f4e4e1
                               at /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/std/src/sys_common/backtrace.rs:65:5
   3:     0x562d63bf4351 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h113657117676131e
                               at /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x562d63c1995f - core::fmt::rt::Argument::fmt::hd56cdfa11c364505
                               at /rustc/2f6bc5d259e7ab25ddfdd33de53b892770218918/library/core/src/fmt/rt.rs:138:9
   5:     0x562d63c1995f - core::fmt::write::h24c20284e5d6be9e
[...]

PR #81, opened 2023-05-11
Fix alignment issues (and a test-only buffer overflow)

saethlin:
Fixes https://github.com/qmonnet/rbpf/issues/77

I ran this to find all the issues fixed here:
```
RUSTDOCFLAGS=-Zsanitizer=address ASAN_OPTIONS=detect_leaks=0 RUSTFLAGS=-Zsanitizer=address cargo +nightly test --target=x86_64-unknown-linux-gnu
```

ASan is still grumpy because `JitMemory` leaks the `contents` allocation. This is probably fine in practice I think because there won't be more than one JIT in a process? But if you want you could also add a `Drop` impl.

It would probably be good to use the above invocation when doing stuff in CI, but I'm awful with GitHub actions so I'll leave that to you if you want to do it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,9 +34,3 @@ byteorder = "1.2"
 elf = "0.0.10"
 json = "0.11"
 hex = "0.4.3"
-
-[profile.dev]
-
-# TODO: Find a way to re-enable debug assertions by addressing the reports for
-# misaligned pointer dereferences - see issue #77.
-debug-assertions = false
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -203,8 +203,8 @@ pub fn sqrti (arg1: u64, unused2: u64, unused3: u64, unused4: u64, unused5: u64)
 /// ```
 /// use rbpf::helpers;
 ///
-/// let foo = "This is a string.".as_ptr() as u64;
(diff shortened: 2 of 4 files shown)


qmonnet (approved):
Looks great, thanks a ton for fixing these!

qmonnet:
> ASan is still grumpy because `JitMemory` leaks the `contents` allocation.

How do you observe these? I don't see any remaining complaints when running the command you provided.

> This is probably fine in practice I think because there won't be more than one JIT in a process? But if you want you could also add a `Drop` impl.

Yes, at the moment there would be only one, I haven't looked at multithreading for rbpf at this stage. I guess I could look into the `Drop` in the future.

> It would probably be good to use the above invocation when doing stuff in CI, but I'm awful with GitHub actions so I'll leave that to you if you want to do it.

Sounds good, and it's alright, I should be able to handle that when I can find a moment for it. Thanks!

saethlin:
> How do you observe these? I don't see any remaining complaints when running the command you provided.

The command I provided sets `ASAN_OPTIONS=detect_leaks=0` so it turns off leak reports. Delete that from the command and you should see complaints about leaks.

qmonnet:
Fixed the memory leak and added flags to CI in #82, thanks again.

PR #84, opened 2023-06-04
bump combine version

yihuaf:
Not sure if you are willing to accept this PR, but nonetheless since I already did it. Currently, this crate uses `combine` crate version 2.5, which is a very old version. The latest version is `4.6`. 

I understand that using an older version by itself is not an issue. I decided to go down this rabbit hole because I noticed the following security notification from dependabot on one of the project I am working on.

```
Dependabot cannot update ascii to a non-vulnerable version
The latest possible version of ascii that can be installed is 0.7.1.

The earliest fixed version is 0.9.3.
```

The older version of `combine` uses the `ascii` crate. However, the latest version of `combine` version 4.6 no longer has this dependency.

Since the major version changed from 2 -> 4, there are breaking changes, which I addressed in this PR. The latest version of `combine` implemented the `display` trait for the parser error, so technically we no longer need to format errors ourselves. However, the format is different, so I decided to keep the error format. I am happy to remove these functions if you wish.

With that being said, I had to make changes to the `test_error_unexpected_character` error messages. The error will container the correct `expected end of input`, but also contain a number of other errors from previous `many` parser. I filed the issue: https://github.com/Marwes/combine/issues/355 to clarify with the `combine` author. As far as I can tell, this is harmless behavior that doesn't impact the correctness of the parsing logic.

There are two commits. The first are `cargo fmt` and the second contains the actual change.

### DIFF
--- a/src/asm_parser.rs
+++ b/src/asm_parser.rs
@@ -5,11 +5,12 @@
 // file of the crate. Do not expect to find those comments in the documentation of the crate.
 
 //! This module parses eBPF assembly language source code.
-
 use combine::char::{alpha_num, char, digit, hex_digit, spaces, string};
-use combine::{between, eof, many, many1, one_of, optional, Parser, ParseError, ParseResult, parser,
-              sep_by, try, State, Stream};
 use combine::primitives::{Error, Info};
+use combine::{
+    between, eof, many, many1, one_of, optional, parser, sep_by, try, ParseError, ParseResult,
+    Parser, State, Stream,
+};
 
 /// Operand of an instruction.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
@@ -34,13 +35,15 @@ pub struct Instruction {
 }
 
 fn ident<I>(input: I) -> ParseResult<String, I>
-    where I: Stream<Item = char>
+where
+    I: Stream<Item = char>,
 {
     many1(alpha_num()).parse_stream(input)
 }
 
 fn integer<I>(input: I) -> ParseResult<i64, I>
-    where I: Stream<Item = char>
+where
+    I: Stream<Item = char>,
 {
     let sign = optional(one_of("-+".chars())).map(|x| match x {
         Some('-') => -1,
@@ -50,11 +53,14 @@ fn integer<I>(input: I) -> ParseResult<i64, I>
         .with(many1(hex_digit()))
         .map(|x: String| u64::from_str_radix(&x, 16).unwrap() as i64);
     let dec = many1(digit()).map(|x: String| x.parse::<i64>().unwrap());
-    (sign, try(hex).or(dec)).map(|(s, x)| s * x).parse_stream(input)
+    (sign, try(hex).or(dec))
+        .map(|(s, x)| s * x)
+        .parse_stream(input)
 }
 
 fn register<I>(input: I) -> ParseResult<i64, I>
-    where I: Stream<Item = char>
+where
+    I: Stream<Item = char>,
 {
     char('r')
         .with(many1(digit()))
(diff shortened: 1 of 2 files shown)


qmonnet:
Thanks for this!

> Not sure if you are willing to accept this PR, but nonetheless since I already did it. Currently, this crate uses `combine` crate version 2.5, which is a very old version. The latest version is `4.6`.
> 
> I understand that using an older version by itself is not an issue. I decided to go down this rabbit hole because I noticed the following security notification from dependabot on one of the project I am working on.
> 
> ```
> Dependabot cannot update ascii to a non-vulnerable version
> The latest possible version of ascii that can be installed is 0.7.1.
> 
> The earliest fixed version is 0.9.3.
> ```
> 
> The older version of `combine` uses the `ascii` crate. However, the latest version of `combine` version 4.6 no longer has this dependency.

Yes, rbpf uses an old version. I haven't seen much interest in updating the dependencies so far, but if it raises security alerts, updates are welcome!

I'm just a bit surprised, because Dependabot raises no alert on `ascii` on rbpf's repo, I would expect it to report it the same way it did for you?

> 
> Since the major version changed from 2 -> 4, there are breaking changes, which I addressed in this PR.

Awesome, thanks! CI tests are all passing so it looks all good from my side.

> The latest version of `combine` implemented the `display` trait for the parser error, so technically we no longer need to format errors ourselves. However, the format is different, so I decided to keep the error format. I am happy to remove these functions if you wish.

I don't think we mind much the format for those, if it's already implemented in the crate I think we could just use that and avoid keeping another version in rbpf. Would you have a sample output for the two versions, by any chance?

> 
> With that being said, I had to make changes to the `test_error_unexpected_character` error messages. The error will container the correct `expected end of input`, but also contain a number of other errors from previous `many` parser. I filed the issue: [Marwes/combine#355](https://github.com/Marwes/combine/issues/355) to clarify with the `combine` author. As far as I can tell, this is harmless behavior that doesn't impact the correctness of the parsing logic.

Should not matter much either, but thanks a lot for raising this and let's see what they say.

> 
> There are two commits. The first are `cargo fmt` and the second contains the actual change.

:+1:

src/asm_parser.rs
```diff
@@ -5,11 +5,12 @@
 // file of the crate. Do not expect to find those comments in the documentation of the crate.
 
 //! This module parses eBPF assembly language source code.
-
```
qmonnet, line 8: Nit: `cargo fmt` doesn't seem to remove this blank line for me, and I'd avoid removing it because the comment is not related to the imports below.

qmonnet (commented):
Any chance we could have a bit more motivation and context in the commit log (mostly for the 2nd commit) please? I find it easier to check out git logs rather than GitHub PR discussions to understand where a change comes from.

yihuaf:
> Any chance we could have a bit more motivation and context in the commit log (mostly for the 2nd commit) please? I find it easier to check out git logs rather than GitHub PR discussions to understand where a change comes from.

Will do :) I will update the PR later tonight or tomorrow.

yihuaf:
> Would you have a sample output for the two versions, by any chance?

Take the `exit\n^` in the test case for example, the current error messages are:

```
Err("Parse error at line 2 column 1: unexpected '^', expected letter or digit, expected whitespaces, expected 'r', expected '-', expected '+', expected '[', expected end of input")`'
```

The `combine` implemented display trait:

```
Err("Parse error at line: 2, column: 1\nUnexpected `^`\nExpected letter or digit, whitespaces, `r`, `-`, `+`, `[` or end of input\n")`
```

The output are very similar, so if you don't mind the actual format of the message, I would propose we just use the `to_string` method instead. I added the change as a different commit for you to consider.

yihuaf:
PTAL

qmonnet (approved):
Looks all good, thanks a lot!

> The output are very similar, so if you don't mind the actual format of the message, I would propose we just use the to_string method instead. I added the change as a different commit for you to consider.

Definitely, let's get rid of this if `combine` offers it already.

Issue #92, opened 2023-12-30
Attempt to negate with overflow in ld_reg_str of disassembler


pcy190:
The disassembler would panic in `ld_reg_str` when it tries to negate the 0x8000i16 value in `-insn.off`

https://github.com/qmonnet/rbpf/blob/4812c52fe1009ee1e3c2307662d175bdd07b5c71/src/disassembler.rs#L37-L42

The PoC program to reproduce:
```
disassembler::disassemble(&[113, 1, 0, 128, 0, 0, 31, 145])
```

This would encounter the following panic:
```
thread '<unnamed>' panicked at 'attempt to negate with overflow'
```

Noted that, this is issue is also encountered in `st_reg_str`, `jmp_imm_str`, `jmp_reg_str` functions.

To enhance the robustness of the disassembler, the negation logic of i16 could be restructured in those function.

qmonnet:
Addressed in https://github.com/qmonnet/rbpf/pull/93.

PR #93, opened 2023-12-30
Fix negation overflow in disassembler

### LINKED ISSUE #91: Attempt to negate with overflow in ld_st_imm_str of disassembler
The disassembler would panic in `ld_st_imm_str` when it tries to negate the 0x8000i16 value in

https://github.com/qmonnet/rbpf/blob/4812c52fe1009ee1e3c2307662d175bdd07b5c71/src/disassembler.rs#L29-L33

The PoC program to reproduce:
```
disassembler::disassemble(&[98, 1, 0, 128, 0, 0, 31, 145])
```

The would panic the disassembler:
```
thread '<unnamed>' panicked at 'attempt to negate with overflow', /rbpf-0.2.0/src/disassembler.rs:33:56
```

To enhance the robustness of the disassembler, the negation logic of i16 could be restructured in `ld_st_imm_str` function.

pcy190:
Fixes #91 and #92


qmonnet (approved):
Looks good, thanks a lot!

I added a non-regression test here as well, can you please sign-off your commit and update again?

PR #97, opened 2024-01-03
Fix nullness check to src value in MOD32 and DIV32

### LINKED ISSUE #95: Attempt to calculate the remainder with a divisor of zero for MOD32 in interpreter
# Details 

In the implementation of `MOD32`, it checks whether the 64-bit source value  is zero. However, during the modulus computation, the source is casted to 32-bit, which could still be zero, e.g., 0xffff0000. According to the kernel ebpf standard, we should check the `modulo by zero` by checking the nullness of the 32-bit source value.

https://github.com/torvalds/linux/blob/610a9b8f49fbcf1100716370d3b5f6f884a2835a/Documentation/bpf/standardization/instruction-set.rst?plain=1#L246-L251

> If execution would result in modulo by zero, for ``BPF_ALU64`` the value of
the destination register is unchanged whereas for ``BPF_ALU`` the upper
32 bits of the destination register are zeroed.

The current implementation of `MOD32_IMM` and `MOD32_REG` lacks the check to the lower 32 bits of source value.
https://github.com/qmonnet/rbpf/blob/4812c52fe1009ee1e3c2307662d175bdd07b5c71/src/interpreter.rs#L230-L233

The following PoC program would cause the `attempt to calculate the remainder with a divisor of zero` panic in the interpreter:
```
lddw r1, 0x100000000
mod32 r0, r1
exit
```

Result:
```
attempt to calculate the remainder with a divisor of zero
thread '' panicked at 'attempt to calculate the remainder with a divisor of zero', src/interpreter.rs:235:47
stack backtrace:
...
   3: rbpf::interpreter::execute_program
             at ./src/interpreter.rs:235:47
   4: rbpf::EbpfVmMbuff::execute_program
             at ./src/lib.rs:300:9
   5: rbpf::EbpfVmRaw::execute_program
             at ./src/lib.rs:1221:9
   6: rbpf::EbpfVmNoData::execute_program
             at ./src/lib.rs:1581:9
```

# Suggestion
Check nullness of 32-bit casted source value before modulus

### LINKED ISSUE #96: Attempt to divide by zero in DIV32 of interpreter
# Details 

In the implementation of `DIV32`, it checks whether the 64-bit source value  is zero. However, during the division computation, the source is casted to 32-bit, which could still be zero, e.g., 0xffff000000000000. According to the kernel ebpf standard, we should check the `division by zero` by checking the nullness of the 32-bit source value.

https://github.com/torvalds/linux/blob/610a9b8f49fbcf1100716370d3b5f6f884a2835a/Documentation/bpf/standardization/instruction-set.rst?plain=1#L246-L251

> If BPF program execution would result in division by zero, the destination register is instead set to zero.

The current implementation of `DIV32_IMM` and `DIV32_REG` lacks the check to the lower 32 bits of source value.

https://github.com/qmonnet/rbpf/blob/4812c52fe1009ee1e3c2307662d175bdd07b5c71/src/interpreter.rs#L217-L220

The following PoC program would cause the `attempt to calculate the remainder with a divisor of zero` panic in the interpreter:
```
lddw r1, 0x100000000
div32 r0, r1
exit
```

Result:
```
attempt to divide by zero
thread '' panicked at 'attempt to divide by zero', src/interpreter.rs:222:45
stack backtrace:
...
   3: rbpf::interpreter::execute_program
             at ./src/interpreter.rs:222:45
   4: rbpf::EbpfVmMbuff::execute_program
             at ./src/lib.rs:300:9
   5: rbpf::EbpfVmRaw::execute_program
             at ./src/lib.rs:1221:9
```

The expected result is that the program executed with r0 set to zero, without panic/error.

# Suggestion
Check nullness of 32-bit casted source value before division

pcy190:
Fixes #95 
Fixes #96


qmonnet (approved):
Looks good, many thanks!

Like for the other PR, I added non-regression tests. Can you please sign-off your commit and update the PR again?

pcy190:
Updated the patch.

PR #102, opened 2024-03-13
src/interpreter.rs: Fix arithmetic left/right shift implementation (mask offset)

### LINKED ISSUE #101: Inconsistences in arithmetic shift implementation (mask offset)
This issue follows #99 and #100. The previous PR demonstrates the implementation incompliance on the logic shift implementation. However, the arithmetic shift operation still meets this problem, and needs the additional mask offset as well.

The current implementation of ARSH64 is not compliant, and undefined behavior happens if we overflow the number of bits we have when trying to shift.

The following PoC program could trigger this inconsistency:

```
mov64 r8, 0x054545ff
arsh64 r8, r8
exit
```

Note that in the kernel interpreter implementation, it masks the `SRC`/`IMM` as well.
https://github.com/torvalds/linux/blob/610a9b8f49fbcf1100716370d3b5f6f884a2835a/kernel/bpf/core.c#L1794-L1805

Hence we need to mask the shift offset with 63 in the following code. Since we don't perform verifier check on the immediate value, we need to mask the `IMM` as well.

https://github.com/qmonnet/rbpf/blob/7bebff57bec037cab5acd57ddbd72363fb163faf/src/interpreter.rs#L289-L290

pcy190:
Fix #101 

We mask the mask offset for arithmetic shifts operaitons. The current version of the BPF Instruction Set Specification specifies that "Shift operations use a mask of 0x3F (63) for 64-bit operations and 0x1F (31) for 32-bit operations" [0].

[0]: https://www.ietf.org/archive/id/draft-thaler-bpf-isa-02.html#section-3.1-2.2.16.1.1

### DIFF
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -286,8 +286,8 @@ pub fn execute_program(prog_: Option<&[u8]>, mem: &[u8], mbuff: &[u8], helpers: 
             ebpf::XOR64_REG  => reg[_dst] ^= reg[_src],
             ebpf::MOV64_IMM  => reg[_dst] =  insn.imm  as u64,
             ebpf::MOV64_REG  => reg[_dst] =  reg[_src],
-            ebpf::ARSH64_IMM => reg[_dst] = (reg[_dst] as i64 >> insn.imm)  as u64,
-            ebpf::ARSH64_REG => reg[_dst] = (reg[_dst] as i64 >> reg[_src]) as u64,
+            ebpf::ARSH64_IMM => reg[_dst] = (reg[_dst] as i64 >> (insn.imm as u64 & SHIFT_MASK_64))  as u64,
+            ebpf::ARSH64_REG => reg[_dst] = (reg[_dst] as i64 >> (reg[_src] as u64 & SHIFT_MASK_64)) as u64,
 
             // BPF_JMP class
             // TODO: check this actually works as expected for signed / unsigned ops


pcy190:
Seems that the casting should be rewritten.

src/interpreter.rs
```diff
@@ -239,8 +239,8 @@ pub fn execute_program(prog_: Option<&[u8]>, mem: &[u8], mbuff: &[u8], helpers:
             ebpf::XOR32_REG  =>   reg[_dst] = (reg[_dst] as u32             ^ reg[_src] as u32) as u64,
             ebpf::MOV32_IMM  =>   reg[_dst] = insn.imm   as u32                                 as u64,
             ebpf::MOV32_REG  =>   reg[_dst] = (reg[_src] as u32)                                as u64,
-            ebpf::ARSH32_IMM => { reg[_dst] = (reg[_dst] as i32).wrapping_shr(insn.imm  as u32) as u64; reg[_dst] &= U32MAX; },
-            ebpf::ARSH32_REG => { reg[_dst] = (reg[_dst] as i32).wrapping_shr(reg[_src] as u32) as u64; reg[_dst] &= U32MAX; },
+            ebpf::ARSH32_IMM => { reg[_dst] = (reg[_dst] as i32).wrapping_shr(insn.imm  as u32 & SHIFT_MASK_32) as u64; reg[_dst] &= U32MAX; },
+            ebpf::ARSH32_REG => { reg[_dst] = (reg[_dst] as i32).wrapping_shr(reg[_src] as u32 & SHIFT_MASK_32) as u64; reg[_dst] &= U32MAX; },
```
qmonnet, line 243: Did you check that this part is necessary? [I wrote some tests](https://github.com/qmonnet/rbpf/tree/arsh-fix-tests) and the 32-bit versions pass successfully without this change. I suspect that the `.wrapping_shr()` does something similar already? If that's the case, maybe just add a comment to explain why we don't need additional masking?

src/interpreter.rs
```diff
@@ -286,8 +286,8 @@ pub fn execute_program(prog_: Option<&[u8]>, mem: &[u8], mbuff: &[u8], helpers:
             ebpf::XOR64_REG  => reg[_dst] ^= reg[_src],
             ebpf::MOV64_IMM  => reg[_dst] =  insn.imm  as u64,
             ebpf::MOV64_REG  => reg[_dst] =  reg[_src],
-            ebpf::ARSH64_IMM => reg[_dst] = (reg[_dst] as i64 >> insn.imm)  as u64,
-            ebpf::ARSH64_REG => reg[_dst] = (reg[_dst] as i64 >> reg[_src]) as u64,
+            ebpf::ARSH64_IMM => reg[_dst] = (reg[_dst] as i64 >> (insn.imm as u64 & SHIFT_MASK_64))  as u64,
+            ebpf::ARSH64_REG => reg[_dst] = (reg[_dst] as i64 >> (reg[_src] as u64 & SHIFT_MASK_64)) as u64,
```
qmonnet, line 290: This looks correct, thanks!

qmonnet (changes requested):
We need some tests coming with this patch, [I wrote some here](https://github.com/qmonnet/rbpf/commit/2cb24f7dcb42f7c1fcb869ead011fc57332636a6) (feel free to add on top of your PR). Based on this tests, I've got one comment inline below.

I bumped the version in the Appveyor workflow, so it should pass if you rebase your PR.

qmonnet:
And thanks a lot!!

src/interpreter.rs
pcy190: Many thanks! This mask is unnecessary since the wrapper_shr has already been implemented as `self >> mask(rhs)` according to the rust docs.

qmonnet (approved):

PR #105, opened 2024-03-15
Follow-up for masks on shift operations

qmonnet:
- tests/ubpf_vm.rs: Add tests for arsh offset masking
- src/interpreter: Remove/explain masks for 32-bit shift operations

Cc: @pcy190


pcy190:
Many thanks for the follow-up!

PR #106, opened 2024-03-18
Add no_std compatibility

SzymonKubica:
### What was added?

Added support for rbpf to run on `no_std` environments. 

### Why is it needed?

Using rbpf on targets without `std` support increases compatibility of the crate. 
I've been using a modified version of rbpf to run eBPF programs on microcontrollers (see [here(https://github.com/SzymonKubica/mibpf))

### Testing

Currently the code builds successfuly with both configuration options enabled, i.e. all build configurations succeed:
```
cargo build
cargo build --features cranelift
cargo build --no-default-features
cargo build --features cranelift --no-default-features
```

All tests (except for the ones using the `print` - only available in std) pass in both configuration options.

### TODO items for the  PR

- [x] test `no_std` on an example embedded device

### Notes

The structure of `std` / `no_std` feature-gating is inspired by `serde`, inside of `lib.rs` a public module: `lib` is defined and it 
contains reexports of all required items from std/alloc/core. It is then used throughout the codebase by importing all items 
declared inside of it: 
```
use crate::lib::*;
```
This behaves similarly as if we were using the standard library as we don't have to import things like `String` or `Vec` manually.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,8 @@ json = "0.11"
 hex = "0.4.3"
 
 [features]
-default = []
+default = ["std"]
+std = []
 cranelift = [
     "dep:cranelift-codegen",
     "dep:cranelift-frontend",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,10 +24,15 @@ include = [
 
 [dependencies]
 
-combine = "4.6"
-libc = "0.2"
-time = "0.2"
-byteorder = "1.2"
+# Default features (std) are disabled so that the dependencies don't pull in the
+# standard library when the crate is compiled for no_std
+byteorder = { version = "1.2", default-features = false }
+log = {version = "0.4.21", default-features = false }
+combine = { version = "4.6", default-features = false }
+
+# Optional Dependencies when using the standard library
+libc = { version = "0.2", optional = true }
+time = { version = "0.2", optional = true }
 
 # Optional Dependencies for the CraneLift JIT
 cranelift-codegen = { version = "0.99", optional = true }
@@ -44,7 +49,8 @@ hex = "0.4.3"
 
 [features]
 default = ["std"]
-std = []
+cargo-clippy = []
+std = ["dep:time", "dep:libc", "combine/std"]
 cranelift = [
     "dep:cranelift-codegen",
     "dep:cranelift-frontend",
@@ -52,3 +58,23 @@ cranelift = [
     "dep:cranelift-native",
     "dep:cranelift-module",
 ]
+
+# Examples that depend on the standard library should be disabled when
+# testing the `no_std` configuration.
+[[example]]
+name = "disassemble"
+required-features = ["std"]
+
+[[example]]
+name = "load_elf"
+required-features = ["std"]
+
+[[example]]
+name = "uptime"
+required-features = ["std"]
+
+[[example]]
+name = "to_json"
+
+[[example]]
+name = "rbpf_plugin"
--- a/README.md
+++ b/README.md
@@ -447,9 +447,12 @@ fn main() {
     let mut vm = rbpf::EbpfVmFixedMbuff::new(Some(prog), 0x40, 0x50).unwrap();
 
     // We register a helper function, that can be called by the program, into
-    // the VM.
-    vm.register_helper(helpers::BPF_TRACE_PRINTK_IDX,
-                       helpers::bpf_trace_printf).unwrap();
+    // the VM. The `bpf_trace_printf` is only available when we have access to
+    // the standard library.
+    #[cfg(feature = "std")] {
+        vm.register_helper(helpers::BPF_TRACE_PRINTK_IDX,
+                           helpers::bpf_trace_printf).unwrap();
+    }
 
     // This kind of VM takes a reference to the packet data, but does not need
     // any reference to the metadata buffer: a fixed buffer is handled
@@ -479,7 +482,9 @@ let prog = assemble("add64 r1, 0x605
                      neg64 r2
                      exit").unwrap();
 
-println!("{:?}", prog);
+#[cfg(feature = "std")] {
+    println!("{:?}", prog);
+}
 ```
 
 The above snippet will produce:
--- a/README.md
+++ b/README.md
@@ -306,10 +306,10 @@ fn main() {
     // directly reads from packet data)
     let mut vm = rbpf::EbpfVmRaw::new(Some(prog)).unwrap();
 
-    #[cfg(windows)] {
+    #[cfg(any(windows, not(feature = "std")))] {
         assert_eq!(vm.execute_program(mem).unwrap(), 0x11);
     }
-    #[cfg(not(windows))] {
+    #[cfg(all(not(windows), feature = "std"))] {
         // This time we JIT-compile the program.
         vm.jit_compile().unwrap();
 
@@ -352,10 +352,10 @@ fn main() {
     // This eBPF VM is for program that use a metadata buffer.
     let mut vm = rbpf::EbpfVmMbuff::new(Some(prog)).unwrap();
 
-    #[cfg(windows)] {
+    #[cfg(any(windows, not(feature = "std")))] {
         assert_eq!(vm.execute_program(mem, mbuff).unwrap(), 0x2211);
     }
-    #[cfg(not(windows))] {
+    #[cfg(all(not(windows), feature = "std"))] {
         // Here again we JIT-compile the program.
         vm.jit_compile().unwrap();
 
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Rust (user-space) virtual machine for eBPF
 * [API](#api)
 * [Example uses](#example-uses)
 * [Building eBPF programs](#building-ebpf-programs)
+* [Build Features](#build-features)
 * [Feedback welcome!](#feedback-welcome)
 * [Questions / Answers](#questions--answers)
 * [Caveats](#caveats)
@@ -554,6 +555,30 @@ program.add(Source::Imm, Arch::X64).set_dst(1).set_imm(0x605).push()
 
 Again, please refer to [the source and related tests](src/insn_builder.rs) to
 get more information and examples on how to use it.
+
+## Build features
+
+### `no_std`
+
+The `rbpf` crate has a Cargo feature named "std" that is enabled by default. To
+use `rbpf` in `no_std` environments this feature needs to be disabled. To do
+this, you need to modify your dependency on `rbpf` in Cargo.toml to disable the
+enabled-by-default features.
+
+```toml
+[dependencies]
+rbpf = { version = "1.0", default-features = false }
+```
+
+Note that when using this crate in `no_std` environments, the `jit` module
+isn't available. This is because it depends on functions provided by `libc`
+(`libc::posix_memalign()`, `libc::mprotect()`) which aren't available on
+`no_std`.
+
+The `assembler` module is available, albeit with reduced debugging features. It
+depends on the `combine` crate providing parser combinators. Under `no_std`
+this crate only provides simple parsers which generate less descriptive error
+messages.
 
 ## Feedback welcome!
 
--- a/examples/rbpf_plugin.rs
+++ b/examples/rbpf_plugin.rs
@@ -15,6 +15,7 @@ fn _unwind(a: u64, _b: u64, _c: u64, _d: u64, _e: u64) -> u64
 // It reads the program from stdin.
 fn main() {
     let mut args: Vec<String> = std::env::args().collect();
+    #[allow(unused_mut)] // In no_std the jit variable isn't mutated.
     let mut jit : bool = false;
     let mut cranelift : bool = false;
     let mut program_text = String::new();
--- a/examples/rbpf_plugin.rs
+++ b/examples/rbpf_plugin.rs
@@ -38,11 +38,11 @@ fn main() {
                 return;
             },
             "--jit" => {
-                #[cfg(windows)] {
-                    println!("JIT not supported on Windows");
+                #[cfg(any(windows, not(feature = "std")))] {
+                    println!("JIT not supported");
                     return;
                 }
-                #[cfg(not(windows))] {
+                #[cfg(all(not(windows), feature = "std"))] {
                     jit = true;
                 }
             },
@@ -93,11 +93,11 @@ fn main() {
 
     let result : u64;
     if jit {
-        #[cfg(windows)] {
-            println!("JIT not supported on Windows");
+        #[cfg(any(windows, not(feature = "std")))] {
+            println!("JIT not supported");
             return;
         }
-        #[cfg(not(windows))] {
+        #[cfg(all(not(windows), feature = "std"))] {
             unsafe {
                 vm.jit_compile().unwrap();
                 result = vm.execute_program_jit(&mut memory).unwrap();
--- a/examples/uptime.rs
+++ b/examples/uptime.rs
@@ -52,14 +52,14 @@ fn main() {
 
     let time;
 
-    #[cfg(not(windows))]
+    #[cfg(all(not(windows), feature = "std"))]
     {
         vm.jit_compile().unwrap();
 
         time = unsafe { vm.execute_program_jit().unwrap() };
     }
 
-    #[cfg(windows)]
+    #[cfg(any(windows, not(feature = "std")))]
     {
         time = vm.execute_program().unwrap();
     }
--- a/src/asm_parser.rs
+++ b/src/asm_parser.rs
@@ -8,10 +8,13 @@
 
 use combine::parser::char::{alpha_num, char, digit, hex_digit, spaces, string};
 use combine::stream::position::{self};
+#[cfg(feature = "std")]
+use combine::EasyParser;
 use combine::{
-    attempt, between, eof, many, many1, one_of, optional, sep_by, EasyParser, ParseError, Parser,
-    Stream,
+    attempt, between, eof, many, many1, one_of, optional, sep_by, ParseError, Parser, Stream,
 };
+
+use crate::lib::*;
 
 /// Operand of an instruction.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
@@ -97,19 +100,30 @@ where
 ///
 /// The instructions are not validated and may have invalid names and operand types.
 pub fn parse(input: &str) -> Result<Vec<Instruction>, String> {
-    match spaces()
-        .with(many(instruction()).skip(eof()))
-        .easy_parse(position::Stream::new(input))
+    let mut with = spaces().with(many(instruction()).skip(eof()));
+
+    #[cfg(feature = "std")]
     {
-        Ok((insts, _)) => Ok(insts),
-        Err(err) => Err(err.to_string()),
+        match with.easy_parse(position::Stream::new(input)) {
+            Ok((insts, _)) => Ok(insts),
+            Err(err) => Err(err.to_string()),
+        }
     }
+    #[cfg(not(feature = "std"))]
+    {
+        match with.parse(position::Stream::new(input)) {
+            Ok((insts, _)) => Ok(insts),
+            Err(err) => Err(err.to_string()),
+        }
+    }
+
 }
 
 #[cfg(test)]
 mod tests {
 
     use super::{ident, instruction, integer, operand, parse, register, Instruction, Operand};
+    use crate::lib::*;
     use combine::Parser;
 
     // Unit tests for the different kinds of parsers.
@@ -566,26 +580,49 @@ exit
         );
     }
 
+    /// When running without `std` the `EasyParser` provided by `combine`
+    /// cannot be used. Because of this we need to use the `Parser` and the
+    /// error messages are different.
     #[test]
     fn test_error_eof() {
+        let expected_error;
+        #[cfg(feature = "std")] {
+            expected_error = Err(
+                "Parse error at line: 1, column: 6\nUnexpected end of input\nExpected digit\n".to_string()
+            );
+        }
+        #[cfg(not(feature = "std"))] {
+            expected_error = Err(
+                "unexpected parse".to_string()
+            );
+        }
         // Unexpected end of input in a register name.
         assert_eq!(
             parse("lsh r"),
-            Err(
-                "Parse error at line: 1, column: 6\nUnexpected end of input\nExpected digit\n"
-                    .to_string()
-            )
+            expected_error
         );
     }
 
+    /// When running without `std` the `EasyParser` provided by `combine`
+    /// cannot be used. Because of this we need to use the `Parser` and the
+    /// error messages are different.
     #[test]
     fn test_error_unexpected_character() {
+        let expected_error;
+        #[cfg(feature = "std")] {
+            expected_error = Err(
+                "Parse error at line: 2, column: 1\nUnexpected `^`\nExpected letter or digit, whitespaces, `r`, `-`, `+`, `[` or end of input\n".to_string()
+            );
+        }
+        #[cfg(not(feature = "std"))] {
+            expected_error = Err(
+                "unexpected parse".to_string()
+            );
+        }
         // Unexpected character at end of input.
         assert_eq!(
             parse("exit\n^"),
-            Err(
-                "Parse error at line: 2, column: 1\nUnexpected `^`\nExpected letter or digit, whitespaces, `r`, `-`, `+`, `[` or end of input\n".to_string()
-            )
+            expected_error
         );
     }
 
--- a/src/assembler.rs
+++ b/src/assembler.rs
@@ -6,10 +6,10 @@
 use asm_parser::{Instruction, Operand, parse};
 use ebpf;
 use ebpf::Insn;
-use std::collections::HashMap;
 use self::InstructionType::{AluBinary, AluUnary, LoadAbs, LoadInd, LoadImm, LoadReg, StoreImm,
                             StoreReg, JumpUnconditional, JumpConditional, Call, Endian, NoOperand};
 use asm_parser::Operand::{Integer, Memory, Register, Nil};
+use crate::lib::*;
 
 #[derive(Clone, Copy, Debug, PartialEq)]
 enum InstructionType {
--- a/src/cranelift.rs
+++ b/src/cranelift.rs
@@ -1,12 +1,5 @@
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-use std::{
-    collections::{BTreeMap, HashMap, HashSet},
-    convert::TryInto,
-    io::ErrorKind,
-    mem::ManuallyDrop,
-};
-
 use cranelift_codegen::{
     entity::EntityRef,
     ir::{
@@ -26,6 +19,7 @@ use crate::ebpf::{
     self, Insn, BPF_ALU_OP_MASK, BPF_JEQ, BPF_JGE, BPF_JGT, BPF_JLE, BPF_JLT, BPF_JMP32, BPF_JNE,
     BPF_JSET, BPF_JSGE, BPF_JSGT, BPF_JSLE, BPF_JSLT, BPF_X, STACK_SIZE, BPF_IND,
 };
+use crate::lib::*;
 
 use super::Error;
 
@@ -1197,7 +1191,7 @@ impl CraneliftProgram {
     /// module, since it's not guaranteed to be valid after the module is dropped.
     pub(crate) fn get_main_function(&self) -> JittedFunction {
         let function_ptr = self.module.get_finalized_function(self.main_id);
-        unsafe { std::mem::transmute(function_ptr) }
+        unsafe { mem::transmute(function_ptr) }
     }
 
     /// Execute this module by calling the main function
--- a/src/disassembler.rs
+++ b/src/disassembler.rs
@@ -4,7 +4,12 @@
 //! Functions in this module are used to handle eBPF programs with a higher level representation,
 //! for example to disassemble the code into a human-readable format.
 
+use log::warn;
+#[cfg(not(feature = "std"))]
+use log::info;
+
 use ebpf;
+use crate::lib::*;
 
 #[inline]
 fn alu_imm_str(name: &str, insn: &ebpf::Insn) -> String {
@@ -20,7 +25,7 @@ fn alu_reg_str(name: &str, insn: &ebpf::Insn) -> String {
 fn byteswap_str(name: &str, insn: &ebpf::Insn) -> String {
     match insn.imm {
         16 | 32 | 64 => {},
-        _ => println!("[Disassembler] Warning: Invalid offset value for {name} insn")
+        _ => warn!("[Disassembler] Warning: Invalid offset value for {name} insn")
     }
     format!("{name}{} r{}", insn.imm, insn.dst)
 }
@@ -384,7 +389,14 @@ pub fn to_insn_vec(prog: &[u8]) -> Vec<HLInsn> {
 /// exit
 /// ```
 pub fn disassemble(prog: &[u8]) {
-    for insn in to_insn_vec(prog) {
-        println!("{}", insn.desc);
+    #[cfg(feature = "std")] {
+        for insn in to_insn_vec(prog) {
+            println!("{}", insn.desc);
+        }
+    }
+    #[cfg(not(feature = "std"))] {
+        for insn in to_insn_vec(prog) {
+            info!("{}", insn.desc);
+        }
     }
 }
--- a/src/ebpf.rs
+++ b/src/ebpf.rs
@@ -15,6 +15,7 @@
 //! the list of the operation codes: <https://github.com/iovisor/bpf-docs/blob/master/eBPF.md>
 
 use byteorder::{ByteOrder, LittleEndian};
+use crate::lib::*;
 
 /// Maximum number of instructions in an eBPF program.
 pub const PROG_MAX_INSNS: usize = 1000000;
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -16,9 +16,10 @@
 //! value. Hence some helpers have unused arguments, or return a 0 value in all cases, in order to
 //! respect this convention.
 
+#[cfg(feature = "std")]
 extern crate libc;
 
-use std::u64;
+use crate::lib::*;
 
 // Helpers associated to kernel helpers
 // See also linux/include/uapi/linux/bpf.h in Linux kernel sources.
@@ -47,6 +48,7 @@ pub const BPF_KTIME_GETNS_IDX: u32 = 5;
 #[allow(dead_code)]
 #[allow(unused_variables)]
 #[allow(deprecated)]
+#[cfg(feature = "std")]
 pub fn bpf_time_getns (unused1: u64, unused2: u64, unused3: u64, unused4: u64, unused5: u64) -> u64 {
     time::precise_time_ns()
 }
@@ -94,6 +96,7 @@ pub const BPF_TRACE_PRINTK_IDX: u32 = 6;
 /// program is run.
 #[allow(dead_code)]
 #[allow(unused_variables)]
+#[cfg(feature = "std")]
 pub fn bpf_trace_printf (unused1: u64, unused2: u64, arg3: u64, arg4: u64, arg5: u64) -> u64 {
     println!("bpf_trace_printf: {arg3:#x}, {arg4:#x}, {arg5:#x}");
     let size_arg = | x | {
@@ -191,6 +194,7 @@ pub fn memfrob (ptr: u64, len: u64, unused3: u64, unused4: u64, unused5: u64) ->
 /// ```
 #[allow(dead_code)]
 #[allow(unused_variables)]
+#[cfg(feature = "std")] // sqrt is only available when using `std`
 pub fn sqrti (arg1: u64, unused2: u64, unused3: u64, unused4: u64, unused5: u64) -> u64 {
     (arg1 as f64).sqrt() as u64
 }
@@ -258,6 +262,7 @@ pub fn strcmp (arg1: u64, arg2: u64, arg3: u64, unused4: u64, unused5: u64) -> u
 /// ```
 #[allow(dead_code)]
 #[allow(unused_variables)]
+#[cfg(feature = "std")]
 pub fn rand (min: u64, max: u64, unused3: u64, unused4: u64, unused5: u64) -> u64 {
     let mut n = unsafe {
         (libc::rand() as u64).wrapping_shl(32) + libc::rand() as u64
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -5,10 +5,8 @@
 // Copyright 2016 6WIND S.A. <quentin.monnet@6wind.com>
 //      (Translation to Rust, MetaBuff/multiple classes addition, hashmaps for helpers)
 
-use std::collections::HashMap;
-use std::io::{Error, ErrorKind};
-
 use ebpf;
+use crate::lib::*;
 
 fn check_mem(addr: u64, len: usize, access_type: &str, insn_ptr: usize,
              mbuff: &[u8], mem: &[u8], stack: &[u8]) -> Result<(), Error> {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,10 +26,17 @@
         unreadable_literal
     )
 )]
+// Configures the crate to be `no_std` when `std` feature is disabled.
+#![cfg_attr(not(feature = "std"), no_std)]
 
 extern crate byteorder;
 extern crate combine;
+#[cfg(feature = "std")]
 extern crate time;
+extern crate log;
+
+#[cfg(not(feature = "std"))]
+extern crate alloc;
 
 #[cfg(feature = "cranelift")]
 extern crate cranelift_codegen;
@@ -43,9 +50,7 @@ extern crate cranelift_module;
 extern crate cranelift_native;
 
 use byteorder::{ByteOrder, LittleEndian};
-use std::collections::HashMap;
-use std::io::{Error, ErrorKind};
-use std::u32;
+use crate::lib::*;
 
 mod asm_parser;
 pub mod assembler;
@@ -59,6 +64,60 @@ mod interpreter;
 #[cfg(all(not(windows), feature = "std"))]
 mod jit;
 mod verifier;
+#[cfg(not(feature = "std"))]
+mod no_std_error;
+
+/// Reexports all the types needed from the `std`, `core`, and `alloc`
+/// crates. This avoids elaborate import wrangling having to happen in every
+/// module. Inspired by the design used in `serde`.
+pub mod lib {
+    mod core {
+        #[cfg(not(feature = "std"))]
+        pub use core::*;
+        #[cfg(feature = "std")]
+        pub use std::*;
+    }
+
+    pub use self::core::convert::TryInto;
+    pub use self::core::mem;
+    pub use self::core::mem::ManuallyDrop;
+    pub use self::core::ptr;
+
+    pub use self::core::{u32, u64, f64};
+
+    #[cfg(feature = "std")]
+    pub use std::println;
+
+    #[cfg(not(feature = "std"))]
+    pub use alloc::vec;
+    #[cfg(not(feature = "std"))]
+    pub use alloc::vec::Vec;
+    #[cfg(feature = "std")]
+    pub use std::vec::Vec;
+
+    #[cfg(not(feature = "std"))]
+    pub use alloc::string::{String, ToString};
+    #[cfg(feature = "std")]
+    pub use std::string::{String, ToString};
+
+    // In no_std we cannot use randomness for hashing, thus we need to use
+    // BTree-based implementations of Maps and Sets. The cranelift module uses
+    // BTrees by default, hence we need to expose it twice here.
+    #[cfg(not(feature = "std"))]
+    pub use alloc::collections::{BTreeMap as HashMap, BTreeMap, BTreeSet as HashSet, BTreeSet};
+    #[cfg(feature = "std")]
+    pub use std::collections::{BTreeMap, HashMap, HashSet};
+
+    /// In no_std we use a custom implementation of the error which acts as a
+    /// replacement for the io Error.
+    #[cfg(not(feature = "std"))]
+    pub use crate::no_std_error::{Error, ErrorKind};
+    #[cfg(feature = "std")]
+    pub use std::io::{Error, ErrorKind};
+
+    #[cfg(not(feature = "std"))]
+    pub use alloc::format;
+}
 
 /// eBPF verification function that returns an error if the program does not meet its requirements.
 ///
@@ -189,7 +248,7 @@ impl<'a> EbpfVmMbuff<'a> {
     /// # Examples
     ///
     /// ```
-    /// use std::io::{Error, ErrorKind};
+    /// use rbpf::lib::{Error, ErrorKind};
     /// use rbpf::ebpf;
     ///
     /// // Define a simple verifier function.
@@ -253,6 +312,7 @@ impl<'a> EbpfVmMbuff<'a> {
     /// // Register a helper.
     /// // On running the program this helper will print the content of registers r3, r4 and r5 to
     /// // standard output.
+    /// # #[cfg(feature = "std")]
     /// vm.register_helper(6, helpers::bpf_trace_printf).unwrap();
     /// ```
     pub fn register_helper(&mut self, key: u32, function: Helper) -> Result<(), Error> {
@@ -506,7 +566,7 @@ impl<'a> EbpfVmMbuff<'a> {
         //  in the kernel; anyway the verifier would prevent the use of uninitialized registers).
         //  See `mul_loop` test.
         let mem_ptr = match mem.len() {
-            0 => std::ptr::null_mut(),
+            0 => ptr::null_mut(),
             _ => mem.as_ptr() as *mut u8,
         };
 
@@ -687,7 +747,7 @@ impl<'a> EbpfVmFixedMbuff<'a> {
     /// # Examples
     ///
     /// ```
-    /// use std::io::{Error, ErrorKind};
+    /// use rbpf::lib::{Error, ErrorKind};
     /// use rbpf::ebpf;
     ///
     /// // Define a simple verifier function.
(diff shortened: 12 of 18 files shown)


src/without_std.rs
```diff
@@ -3,7 +3,6 @@ pub mod without_std {
     use alloc::string::String;
     pub use core::u32;
     pub use core::u64;
-
```
qmonnet: Why removing this line?

src/lib.rs
```diff
@@ -64,7 +64,7 @@ mod stdlib {
 
 use byteorder::{ByteOrder, LittleEndian};
 use stdlib::collections::HashMap;
-use stdlib::collections::Vec;
+use stdlib::vec::Vec;
```
qmonnet: If such changes are necessary to “make it build successfully”, they should likely be fold into the previous commit. Again, more descriptive commit logs to understand what happens would be nice :)

examples/rbpf_plugin.rs
```diff
@@ -42,7 +42,7 @@ fn main() {
                     println!("JIT not supported on Windows");
                     return;
                 }
-                #[cfg(jit)] {
+                #[cfg(feature = "jit")] {
```
qmonnet: Please squash these with the first commit.

Cargo.toml
```diff
@@ -43,7 +43,8 @@ json = "0.11"
 hex = "0.4.3"
 
 [features]
-default = ["std"]
+default = ["std", "jit"]
+jit = []
```
qmonnet, line 47: This would also need to go into the first commit.

It's still subject to the caveats regarding support for Windows.

SzymonKubica:
Hi, 
Thank you for reviewing the PR and listing the required changes. 
I will start going through these and clarify in the comments above if anything turns out to be unclear. 

No worries about the delay, my goal with this PR was to contribute back a small subset of changes that I needed to introduce to 
rbpf for my project. I've been working on a heavily-modified fork for a while and added an ARMv7 JIT to it. Given the scope of those changes and the fact that they are coupled with my specific requirements, I don't think I will be trying to make them mergeable with the upstream rbpf (although if you want to have a look it's [here](https://github.com/SzymonKubica/rbpf-for-microcontrollers/tree/7cc7b52b479102c4c41ab3abe411ba3de7633414))

I think compared to those changes, adding no_std support is potentially useful so I will continue working on this PR

src/without_std.rs
```diff
@@ -0,0 +1,35 @@
+/// Ensures compatibility with no_std.
+pub mod without_std {
+    use alloc::string::String;
+    pub use core::u32;
+    pub use core::u64;
+
+    // Adapt this with whatever printing functionality your host OS requires.
```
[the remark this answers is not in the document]
SzymonKubica: Here the idea is that in no_std there is no println! macro available by default. Different environments provide their own versions of that macro. For instance, for my use-case, I'm using Rust with RIOT OS and they provide a separate print
macro that allows for printing things to the shell available over the serial port. The idea here is that the print functionality should be overridden with the one that users have access to in their no_std environment. 

This is not a copy-paste, however I agree that it is a somewhat hacky way of allowing to use println in no_std. I will try
to come up with something better. Here are all of the current usages of println macro:
image

They can be divided into three categories:
- print statements used in example files
- printing used as logging (e.g. in the disassembler)
- printing used inside of a helper function (`bpf_trace_printf`)

What I suggest that we do:
- leave example files as they are but ensure that they only run with std enabled
- for the logging use-cases, I suggest that we replace those calls with usages of the logging macros from the `log` crate
- for the printing helper functions, I think we can feature-gate those to be only available with std.

That way we don't have to have this hacky override that needs to be set by the users. 
The reason for using `log` is that in no_std it is able to correctly identify the printing macro that is exposed by the platform that we are running on (even if it has no standard println! macro) For instance in RIOT calling the log macros correctly prints to the shell available over the serial port.

examples/rbpf_plugin.rs
SzymonKubica: Ok that's done, would you say I should keep all changes that I add with my PR within a single commit so that it is easier to review? It is currently done like this

src/ebpf.rs
```diff
@@ -15,6 +15,7 @@
 //! the list of the operation codes: <https://github.com/iovisor/bpf-docs/blob/master/eBPF.md>
 
 use byteorder::{ByteOrder, LittleEndian};
+use crate::stdlib::vec::Vec;
```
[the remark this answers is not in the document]
SzymonKubica: In order to use Vec in no_std we need to explicitly import it. However now that I look at this setup with different modules:
with_alloc, with_std, without_str it looks a bit too cluttered. I had a look at other crates and how they are doing no_std/std conditional switching and found an example from serde [here](https://docs.rs/serde/latest/src/serde/lib.rs.html#95:~:text=///%20A%20facade%20around,num%3A%3ASaturating%3B%0A%7D). It effectively does the same thing as the three of my with_* modules do but in a much cleaner way. I think I will reimplement the setup that there are using

examples/rbpf_plugin.rs
qmonnet: No, please don't do a single commit for all the PR! Some of the steps are logically separated, like adding the `jit` feature flag and then handling the changes for no_std support.

What I would like is a set of logical, but clean commits. In a sense that each commit should contain a logical change, _and_ should be self-contained. There shouldn't be later commit with fixes for the first commits; instead, all “fixes” should be squashed into the commits they improve, to make sure each commit works on its own, and doesn't break the build.

Does it make sense?

SzymonKubica:
One note regarding your comment about asm_parser and disassembler being disabled:
I have looked into those and I don't see anything which would prevent them from running under no_std. I will update those so that they don't have to be disabled. It was probably an oversight on my part as I was primarily interested in using the VM.

SzymonKubica:
Ok I did some restructuring and now the import switching is done the same way serde does it.
I also changed some dependencies and now all modules build with both std/no_std.\
I will look into the jit feature and see if I can make it work with windows (it could be that I can make the  JIT enabled in no_std, in which case I could simply revert to using the windows feature flag)

examples/rbpf_plugin.rs
SzymonKubica: Yes that makes sense I will do my best to follow this practice. Thank you for the feedback, that's really helpful and I agree that it would make it much easier to review and then go through later on.

qmonnet: You're welcome. Kernel docs give more guidance about this, if you wanted more:
- https://docs.kernel.org/process/submitting-patches.html#separate-your-changes
- https://docs.kernel.org/process/5.Posting.html#patch-preparation

The objective is to be able to review logical changes independently, and efficiently. One motivation to preserve the builds through each commit is to be able to `git bisect` efficiently, if needed.

I'll review the rest of your changes soon. Thanks a lot!

examples/rbpf_plugin.rs
```diff
@@ -42,7 +42,7 @@ fn main() {
                     println!("JIT not supported on Windows");
                     return;
                 }
-                #[cfg(not(windows))] {
+                #[cfg(jit)] {
```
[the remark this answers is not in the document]
SzymonKubica: I think I might have a simpler solution. If we consider the possible cases, we are either:
- running on linux with std (in which case we want the jit enabled by default)
- running on windows (jit currently disabled)
- running with no_std (we want jit to be disabled as it requires the posix functions from libc for the jitted program buffer)
We can capture the above cases by enabling the jit only when both `feature = "std"` and `not(windows)` and disabling it when we are either on windows or no_std (or both), I'm currently doing the conditional switching as follows:
```
    #[cfg(any(not(feature = "std"), windows))] {
        assert_eq!(vm.execute_program(mem, mbuff).unwrap(), 0x2211);
    }
    #[cfg(all(feature = "std", not(windows)))] {
        // Here again we JIT-compile the program.
        vm.jit_compile().unwrap();

        // Here we must provide both a reference to the packet data, and to the
        // metadata buffer we use.
        unsafe {
            assert_eq!(vm.execute_program_jit(mem, mbuff).unwrap(), 0x2211);
        }
    }

```

examples/rbpf_plugin.rs
SzymonKubica: Ok thank you for the additional pointers, I've read through those, that's really useful.
Now that I think about this, the jit flag complicates things quite a bit and I think I can get around that using combinations of cfg not(windows) and feature = "std" (see replies above). In this case I think I will open up a new PR and apply the current changes there in a clean way in a single commit. The thing with adding no_std is that I can't really do it half way through so (it needs to be 'atomic') otherwise it won't compile without std. In which case I think it is a single logical change and could be put into a single commit. 

Thank you for all of your comments and suggestions in this thread. Is it ok if I open the new clean PR and have the final version of changes there without all of this intermediate changes that happened here? Or should I continue here so that our conversation gets preserved?

README.md
```diff
@@ -688,8 +688,8 @@ details.
   such as in `mov r0, 0x1337`), by Rich Lane for Big Switch Networks (2015)
 
 * [_Building a simple JIT in
-  Rust_](http://www.jonathanturner.org/2015/12/building-a-simple-jit-in-rust.html),
-  by Jonathan Turner (2015)
+  Rust_](https://www.sophiajt.com/building-a-simple-jit-in-rust),
+  by Sophia Turner (2015)
```
qmonnet, line 692: This looks like it was picked by mistake while rebasing/merging, it's not related to the PR.

src/ebpf.rs
```diff
@@ -18,7 +18,7 @@ use byteorder::{ByteOrder, LittleEndian};
 use crate::stdlib::vec::Vec;
 
 /// Maximum number of instructions in an eBPF program.
-pub const PROG_MAX_INSNS: usize = 4096;
+pub const PROG_MAX_INSNS: usize = 1000000;
```
qmonnet, line 21: This looks like it was picked by mistake while rebasing/merging, it's not related to the PR.

examples/rbpf_plugin.rs
```diff
@@ -5,8 +5,7 @@
 use std::io::Read;
 
 // Helper function used by https://github.com/Alan-Jowett/bpf_conformance/blob/main/tests/call_unwind_fail.data
-fn _unwind(a: u64, _b: u64, _c: u64, _d: u64, _e: u64) -> u64
-{
+fn _unwind(a: u64, _b: u64, _c: u64, _d: u64, _e: u64) -> u64 {
```
qmonnet, line 8: You've got a number of changes that are unrelated to the rest of the PR in this file. Let's leave formatting fixes out of this, please.

examples/rbpf_plugin.rs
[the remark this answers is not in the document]
qmonnet: Yes, it works as well. It's more verbose in the conditions and less explicit that the `jit` feature, but it certainly boils down to the same. I'm OK with that, we can always look at feature aliases as a follow-up if we want.

src/ebpf.rs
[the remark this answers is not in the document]
qmonnet: OK I was just not sure why we suddenly needed to import it whereas that was not the case so far.

Good thing if you find a cleaner way :+1:

src/without_std.rs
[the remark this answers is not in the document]
qmonnet: I suppose it's OK to use the `log` crate, no objection. The rest of your proposal sounds fine as well.

examples/rbpf_plugin.rs
qmonnet: You're free to open a new PR, but I don't really see the point to be honest, it's perfectly fine to update and force-push here.

I'd still prefer having several, smaller commits. For example, adding the `std` feature to the Cargo.toml and using it to gate the related code in addition to `windows` is a logical change and doesn't break the build, even if the `std` feature has not been implemented yet. I'm pretty sure some other parts can be their own commits, too. Don't lose too much time on that, but generally speaking, just keep in mind that smaller commits are easier to review and understand in the future.

qmonnet:
Can you please [sign-off](https://wiki.linuxfoundation.org/dco) your commits?

examples/rbpf_plugin.rs
SzymonKubica: Ok sounds good, thank you for your answer. I do agree that opening up a new RP could potentially create additional mess. I will force-push the clean changes to this one then and try to break them up into separate logical commits

src/ebpf.rs
SzymonKubica: Yes this was a merge, I will checkout the latest upstream and apply the clean changes there.

examples/rbpf_plugin.rs
SzymonKubica: Sorry about the formatting fixes, my autoformatter must have kicked in, I will make sure that the final changes don't touch the code formatting

SzymonKubica:
Re: signing off commits
Yes I will do that from now on, I think some of my previous commits were already signed-off, but I noticed that the build doesn't go through because the old ones weren't signed-off. I will apply the clean changes by signing off commits and force push the update

SzymonKubica:
Ok @qmonnet I think the PR is in much better state now. Following your advice I have split my changes into two logical commits:
- adding the `"std"` feature and modifying the feature-gating for the jit to require that both `not(windows)` and `feature = "std"` are set.
- adding the proper reexports for `no_std`
I updated the description of the PR, now everything builds correctly and all available tests pass in both std and no_std. 

I have also tested that it builds correctly on an actual no_std target: STM32-f439zi running RIOT. I managed to flash this example program: 
```
#![no_std]

use riot_wrappers::riot_main;
use riot_wrappers::println;
use riot_wrappers::riot_sys::malloc;

extern crate rust_riotmodules;
extern crate rbpf;
extern crate alloc;

riot_main!(main);

fn main() {
    let prog = &[
         0x79, 0x11, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, // Load mem from mbuff into R1.
         0x69, 0x10, 0x02, 0x00, 0x00, 0x00, 0x00, 0x00, // ldhx r1[2], r0
         0x95, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00  // exit
     ];
     let mem = &mut [
         0xaa, 0xbb, 0x11, 0x22, 0xcc, 0xdd
     ];

     // Just for the example we create our metadata buffer from scratch, and we store the
     // pointers to packet data start and end in it.
     let mut mbuff = [0u8; 32];
     unsafe {
         let mut data     = mbuff.as_ptr().offset(8)  as *mut u64;
         let mut data_end = mbuff.as_ptr().offset(24) as *mut u64;
         *data     = mem.as_ptr() as u64;
         *data_end = mem.as_ptr() as u64 + mem.len() as u64;
     }

     // Instantiate a VM.
     let mut vm = rbpf::EbpfVmMbuff::new(Some(prog)).unwrap();

     // Provide both a reference to the packet data, and to the metadata buffer.
     let res = vm.execute_program(mem, &mut mbuff).unwrap();
     assert_eq!(res, 0x2211);
     println!("res: 0x{:x}", res);
}
```
It looks like everything works in `no_std`. One note: I ended up having to disable the `assembler` and `asm_parser` as they depend on the `combine` crate which doesn't have the `no_std` mode. 

I can quickly publish a repo with the experimental setup that I used to test the new version of rbpf running without `std` in RIOT if you would find this helpful.

examples/disassemble.rs
```diff
@@ -1,37 +1,40 @@
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 // Copyright 2017 6WIND S.A. <quentin.monnet@6wind.com>
 
+
```
qmonnet, line 4: No need to add a blank line here.

examples/disassemble.rs
```diff
@@ -1,37 +1,40 @@
...
-        0x5f, 0x21, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
-        0xb7, 0x00, 0x00, 0x00, 0xff, 0xff, 0xff, 0xff,
-        0x18, 0x02, 0x00, 0x00, 0x00, 0x00, 0x99, 0x99,
-        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
-        0x1d, 0x21, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00,
-        0xb7, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
-        0x95, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
-    ];
-    disassembler::disassemble(prog);
+    #[cfg(feature = "std")] {
```
qmonnet, line 10: What happens if we run the example without `std`, nothing? Should there be a message instead, or a way to skip compilation entirely in that case?

examples/load_elf.rs
```diff
@@ -49,85 +49,88 @@ use rbpf::helpers;
...
-        assert_eq!(res, 0);
-    }
-
-    #[cfg(any(windows, not(feature = "std")))]
-    {
-        let res = vm.execute_program(packet2).unwrap();
-        println!("Packet #2, program returned: {:?} ({:#x})", res, res);
-        assert_eq!(res, 0);
+    // This example relies on the bpf_trace_printf which is only available on std
+    #[cfg(feature = "std")] {
```
qmonnet, line 53: Do you know if there's a way to mark the whole file as dependent on the feature, and skip this example entirely if `std` is not selected?

examples/uptime.rs
```diff
@@ -11,65 +11,67 @@ use rbpf::helpers;
...
 // The two eBPF programs are independent and are not related to one another.
 fn main() {
-    let prog1 = &[
-        0xb4, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // mov32 r0, 0
-        0xb4, 0x01, 0x00, 0x00, 0x02, 0x00, 0x00, 0x00, // mov32 r1, 2
-        0x04, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, // add32 r0, 1
-        0x0c, 0x10, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // add32 r0, r1
-        0x95, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00  // exit and return r0
-    ];
+    #[cfg(feature = "std")] {
```
qmonnet, line 14: Same remark here, I'd rather skip building the example entirely if we can't run it rather than indenting all the code here and having nothing happening when `std` is turned off.

src/disassembler.rs
```diff
@@ -4,7 +4,9 @@
 //! Functions in this module are used to handle eBPF programs with a higher level representation,
 //! for example to disassemble the code into a human-readable format.
 
+use crate::lib::*;
 use ebpf;
+use log::warn;
```
qmonnet, line 9: How do you sort these? I'd maybe keep local imports last?

src/helpers.rs
```diff
@@ -250,6 +254,7 @@ pub fn strcmp (arg1: u64, arg2: u64, arg3: u64, unused4: u64, unused5: u64) -> u
 /// extern crate time;
 ///
 /// unsafe {
+///
```
qmonnet, line 257: Why the blank line?

src/interpreter.rs
```diff
@@ -5,8 +5,8 @@
 // Copyright 2016 6WIND S.A. <quentin.monnet@6wind.com>
 //      (Translation to Rust, MetaBuff/multiple classes addition, hashmaps for helpers)
 
-use std::collections::HashMap;
-use std::io::{Error, ErrorKind};
+
```
qmonnet, line 8: What the blank line?

src/lib.rs
```diff
@@ -27,9 +27,17 @@
     )
 )]
 
+#![cfg_attr(not(feature = "std"), no_std)]
+
```
qmonnet, line 31: This one is complex to read, please add a comment. Do we need the blank line? (If it applies only to the byteorder crate, probably not?)

src/lib.rs
```diff
@@ -59,6 +66,63 @@ mod interpreter;
 #[cfg(all(not(windows), feature = "std"))]
 mod jit;
 mod verifier;
+#[cfg(not(feature = "std"))]
+mod no_std_error;
+
+/// A facade around all the types needed from the `std`, `core`, and `alloc`
```
qmonnet, line 72: “facade”: I didn't know it existed without the cedilla! :exploding_head:

src/lib.rs
```diff
@@ -724,37 +789,39 @@ impl<'a> EbpfVmFixedMbuff<'a> {
...
+    ///
+    ///     // Instantiate a VM.
+    ///     let mut vm = rbpf::EbpfVmFixedMbuff::new(Some(prog), 0x40, 0x50).unwrap();
+    ///
+    ///     // Register a helper. This helper will store the result of the square root of r1 into r0.
+    ///     vm.register_helper(1, helpers::sqrti);
+    ///
+    ///     let res = vm.execute_program(mem).unwrap();
+    ///     assert_eq!(res, 3);
+    ///     }
```
qmonnet, line 824: Should be unindented?

```suggestion
    /// }
```

src/no_std_error.rs
```diff
@@ -0,0 +1,35 @@
...
+#[derive(Debug)]
+pub struct Error {
+    #[allow(dead_code)]
+    kind: ErrorKind,
+    #[allow(dead_code)]
+    error: String,
+}
+
+impl Error {
+    /// New function added for compatibility with the existing code.
```
qmonnet, line 19: “the existing code” only makes sense in the context of the PR. After this is merged, it doesn't mean much: everything is existing code. Could you please rephrase?

src/no_std_error.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub fn new<S: Into<String>>(kind: ErrorKind, error: S) -> Error {
+        Error {
+            kind,
+            error: error.into(),
+        }
+    }
+}
+
+/// The minimum set of variants to make the dummy ErrorKind work with
+/// the existing code.
```
qmonnet, line 29: Same comment.

tests/misc.rs
```diff
@@ -17,9 +17,11 @@
 // extern crate elf;
 // use std::path::PathBuf;
 
-extern crate rbpf;
+// These tests should only be run on std as they depend on the assembler.
+#![cfg(feature = "std")]
```
qmonnet, line 21: Do they? It seems a lot of them don't need the assembler. Plus most of them seem already gated behind individual `std` directives?

tests/ubpf_verifier.rs
```diff
@@ -16,11 +16,15 @@
...
 
+// These tests should only be run on std as they depend on the assembler.
+#![cfg(feature = "std")]
+
 extern crate rbpf;
 
 use rbpf::assembler::assemble;
 use rbpf::ebpf;
 
+
```
qmonnet, line 27: No need for that

tests/ubpf_verifier.rs
```diff
@@ -16,11 +16,15 @@
 
 // These are unit tests for the eBPF “verifier”.
 
+// These tests should only be run on std as they depend on the assembler.
+#![cfg(feature = "std")]
```
qmonnet, line 20: Not all of them do, but OK. I'd adjust the comment to say "most" or even "many of these tests", rather than implying they all do.

qmonnet (commented):
Looks much better, thanks! Some comments here and there.

Cargo.toml
```diff
@@ -24,10 +24,13 @@ include = [
 
 [dependencies]
 
-combine = "4.6"
-libc = "0.2"
-time = "0.2"
-byteorder = "1.2"
+byteorder = { version = "1.2", default-features = false }
+log = {version = "0.4.21", default-features = false }
```
qmonnet, line 28: What does `default-features = false` do, please?

qmonnet:
Please clean up the commit log from your second commit, you have multiple sign-off tags (likely because you squashed commits together and kept the descriptions of all commits).

Cargo.toml
SzymonKubica: This is a standard way of telling the dependencies that we are using them with `no_std`. In most cases, the way those libraries are designed is similar to how I specified in `Cargo.toml`: 
```toml
[features]
default = ["std"]
std = ["dep:time", "dep:libc", "dep:combine"]
```
And so if someone was to include `rbpf` in their project and specified: 
```toml
rbpf = { version = "1.2", default-features = false }
```
Then the `std` feature would be disabled and so the crate behaves in a way that is compatible with `no_std`.
```toml
[dependencies]
riot-wrappers = { version = "0.8", features = [ "set_panic_handler" ] }
rbpf = {path = "../../../../vms/rbpf-for-microcontrollers", default-features = false}

# While currently this example does not use any RIOT modules implemented in
# Rust, that may change; it is best practice for any RIOT application that has
# its own top-level Rust crate to include rust_riotmodules from inside
# RIOTBASE.
rust_riotmodules = { path = "../../sys/rust_riotmodules/" }
```
This is an example of how I included `rbpf` without default features in that example program with RIOT on stm32

SzymonKubica: I can add a comment in the `Cargo.toml` explaining why we don't need the default features

examples/load_elf.rs
SzymonKubica: That's probably the best way of doing this, I will try to see if it is possible, so far I've tried using: 
```rust
#![cfg(feature = "std")]
```
At the top level of the file but it seems like the example still gets included and the tests complain about it not having the  main function.

tests/misc.rs
SzymonKubica: Good point, I will enable a more fine-grained control so that everything that can possibly be run runs when testing without std

tests/ubpf_verifier.rs
SzymonKubica: Yes, you are right indeed. It was a mistake on my part, I saw that compilation failed because the module was trying to import the assembler so I just disabled the whole test file. Now it is fixed, the assembler gets included conditionally and only the tests that depend on it are disabled when testing `no_std`

src/asm_parser.rs
```diff
@@ -593,7 +598,7 @@ exit
     fn test_initial_whitespace() {
         assert_eq!(
             parse(
-                " 
+                "
```
SzymonKubica, line 601: Not sure what happened here, I tried bringing the space back but it seems like it gets clipped when I commit (it could be that my editor clipps dangling whitespace) or some linux CRLF / LF issue

SzymonKubica:
Ok I think all of the comments have been addressed. Thank you very much for the feedback. 
Key changes introduced:
- now examples are configured inside `Cargo.toml` and the ones that need std are disabled when testing on `no_std`
- the `misc`, `ubpf_verfier`, `ubpf_vm` tests now run all available tests under `no_std` (no more coarse-grained file level feature-gating) Thank you for picking up on this issue. 
- removed whitespace and formatting issues that were introduced by mistake
- reordered the imports as you suggested
- added required explanation / clarification comments

examples/disassemble.rs
SzymonKubica: Good point, I managed to disable the entire example when testing without `std`

examples/load_elf.rs
SzymonKubica: Turns out `Cargo.toml` lets you specify which features the examples depend on.

src/asm_parser.rs
SzymonKubica: Ok I used a different editor and managed to get this diff to disappear

src/lib.rs
```diff
@@ -42,15 +44,37 @@ extern crate cranelift_module;
...
+use stdlib::{Error, ErrorKind};
 
+#[cfg(std)]
 mod asm_parser;
+#[cfg(std)]
 pub mod assembler;
 #[cfg(feature = "cranelift")]
 mod cranelift;
+#[cfg(std)]
 pub mod disassembler;
```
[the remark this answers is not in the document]
SzymonKubica: Yes, I agree. I'm not sure what would be the best place for this doc. I think I could add a section similar to [this one](https://serde.rs/no-std.html) to the readme. Would this work?

SzymonKubica:
Hi @qmonnet, 
I added a description of the `no_std` feature to the README, it was based on how `serde` explains their implementation of `no_std` support. This was done to resolve one of your eariler comments that it isn't obvious what is available and what isn't under
`no_std`. I think this was the last comment that was left to be addressed, so feel free to have another look at the PR when you're available. Thank you for all of your feedback, it was really helpful to learn how to prepare a patch properly.

src/assembler.rs
```diff
@@ -1,15 +1,17 @@
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 // Copyright 2017 Rich Lane <lanerl@gmail.com>
 
-//! This module translates eBPF assembly language to binary.
+//! This module translates eBPF assembly language to binary. It depends on the
+//! `combine` crate, which depends on the standard library. Because of this
+//! it is only available when the `std` feature is enabled.
```
qmonnet, line 6: Looking at the amount of tests that you had to leave out when we use `std`, it's rather annoying to leave the assembler out. How much did you look into `combine`'s support for no_std? [Looking at its changelog](https://github.com/Marwes/combine/blob/f726b8f943f44d258686911907c0324771a2d09f/CHANGELOG.md?plain=1#L446), it should be able to work without `std` since v3 (we're using v4+). Could we deselect `std` for that crate when we deselect it for rbpf, keep building the assembler (and disassembler if possible), and keep the tests?

Cargo.toml
```diff
@@ -43,11 +48,33 @@ json = "0.11"
...
+[[example]]
+name = "uptime"
+required-features = ["std"]
+
+[[example]]
+name = "to_json"
+
+[[example]]
+name = "rbpf_plugin"
+
```
qmonnet, line 80:
```suggestion
```

README.md
```diff
@@ -19,6 +19,7 @@ Rust (user-space) virtual machine for eBPF
 * [Building eBPF programs](#building-ebpf-programs)
 * [Feedback welcome!](#feedback-welcome)
 * [Questions / Answers](#questions--answers)
+* [No_std Support](#no_std-support)
```
qmonnet, line 22: Let's maybe have a “Build features” section (or something like this) instead, and move it higher up in the list (above “Feedback welcome!” maybe). Then we can have `no_std` as a subsection of that? We need to document the Cranelift JIT at some point (not in this PR of course), and it would be another subsection of “Build features”.

examples/load_elf.rs
```diff
@@ -48,7 +48,6 @@ use rbpf::helpers;
 // Once the bytecode has been edited, we can load the bytecode directly from the ELF object file.
 
 fn main() {
-
```
qmonnet, line 51: Unrelated

tests/misc.rs
```diff
@@ -18,8 +18,8 @@
 // use std::path::PathBuf;
 
 extern crate rbpf;
-
```
qmonnet, line 21: Let's keep an empty line here?

src/assembler.rs
SzymonKubica: Good point, I will try and enable it again tonight. From what I remember, simply specifying `default-features=false` in the Cargo.toml didn't cut it but given the change log that you have shared there might still be a way to use it under no_std. It could be that the alloc feature needs to be enabled. 

I do agree that quite a few tests needed to be left out and it would be good to enable those. 
As for the disassembler, this module seems to work fine with `no_std`, it is just that some of its tests require the assembler to prepare the test case which means that they need to be disabled

tests/ubpf_vm.rs
```diff
@@ -9,7 +9,6 @@
 // <https://github.com/iovisor/ubpf/tree/master/tests> (hyphen had to be replaced with underscores
 // as Rust will not accept them in function names). It is strongly advised to refer to the uBPF
 // version to understand what these program do.
-//
```
qmonnet, line 12: Please keep the line :)

qmonnet (changes requested):
Some small nits

src/assembler.rs
SzymonKubica: The changelog does indeed look promising, I was probably just misconfiguring something. I don't think there is something that needs to happen when rbpf calls into `combine` that would require the standard library. From what I could tell it was some parser combinators stuff

qmonnet:
It could be interesting to add the following change as an additional commit to your PR:

```diff
diff --git a/.github/workflows/test.yaml b/.github/workflows/test.yaml
index 60eefc28d727..a11391c430ca 100644
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -17,10 +17,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        toolchain: [stable, beta, nightly]
+        features: [--all-features]
         include:
-          - toolchain: stable
-          - toolchain: beta
           - toolchain: nightly
+            features: --no-default-features
 
     steps:
       - name: Checkout code
@@ -33,12 +34,12 @@ jobs:
           override: true
           components: clippy
 
-      - name: Build with ${{ matrix.toolchain }}
+      - name: Build with ${{ matrix.toolchain }}, ${{ matrix.features }}
         run: |
           cargo +${{ matrix.toolchain }} build \
-              --release --all-features --all-targets
+              --release ${{ matrix.features }} --all-targets
 
-      - name: Test with ${{ matrix.toolchain }}
+      - name: Test with ${{ matrix.toolchain }}, ${{ matrix.features }}
         run: |
           if [[ "${{ matrix.toolchain }}" == 'nightly' ]]; then
               export RUSTDOCFLAGS='-Zsanitizer=address'
@@ -47,7 +48,7 @@ jobs:
           fi
           cargo +${{ matrix.toolchain }} test \
               --target=x86_64-unknown-linux-gnu \
-              --all-features
+              ${{ matrix.features }}
 
       - name: Lint with ${{ matrix.toolchain }}
         run: |
```

I've not tested it, but if I didn't make any mistake it should add a workflow job to build and run the (available) tests without the `std` feature.

tests/ubpf_vm.rs
SzymonKubica: Yes sorry about that, must have deleted it accidentally and didn't spot it in the diff. It was inserted back.

src/assembler.rs
SzymonKubica: Oh, I doublechecked and unfortunately it seems like `rbpf` depends on the `EasyParser` from combine which for some reason is available on `std` only (see [here](https://docs.rs/combine/latest/combine/trait.EasyParser.html#:~:text=Available%20on%20crate%20feature%20std%20only.)

In this case there are two ways of moving forward: 
- leave `asm_parser` and assembler disabled
- change the implementation of the `asm_parser` to not depend on the `EasyParser`

From what I can tell, there is only one place where `rbpf` depends on that trait:
```rust
/// Parse a string into a list of instructions.
///
/// The instructions are not validated and may have invalid names and operand types.
pub fn parse(input: &str) -> Result<Vec<Instruction>, String> {
    match spaces()
        .with(many(instruction()).skip(eof()))
        .easy_parse(position::Stream::new(input))
    {
        Ok((insts, _)) => Ok(insts),
        Err(err) => Err(err.to_string()),
    }
}
```
According to the docs, the easy parser is supposed to produce better error messages. However in this place the error is turned directly into a string, so the code doesn't actually depend on the structue of the returned errors. I will and see if I can replace that with the normal parser. There are two possible approaches:
- replace this with the regular parser and sacrifice the nice error messagea
- instrument the code to use `EasyParser` when we have access to `std` and `Parser` otherwise

qmonnet: If we end up skipping two tests only, then yes it's absolutely fine! Thanks a lot for looking into this!

SzymonKubica: Yes, it turned out that the only difference between `EasyParser` and the `Parser` in combine is that they produce different error messages. Only these two test cases above depend on the contents of those errors, so I added some conditional `cfg` to change the expedted error depending on the enabled features. Aside from this all other tests behave the same under both environments.

SzymonKubica:
> It could be interesting to add the following change as an additional commit to your PR:
> 
> ```diff
> diff --git a/.github/workflows/test.yaml b/.github/workflows/test.yaml
> index 60eefc28d727..a11391c430ca 100644
> --- a/.github/workflows/test.yaml
> +++ b/.github/workflows/test.yaml
> @@ -17,10 +17,11 @@ jobs:
>      strategy:
>        fail-fast: false
>        matrix:
> +        toolchain: [stable, beta, nightly]
> +        features: [--all-features]
>          include:
> -          - toolchain: stable
> -          - toolchain: beta
>            - toolchain: nightly
> +            features: --no-default-features
>  
>      steps:
>        - name: Checkout code
> @@ -33,12 +34,12 @@ jobs:
>            override: true
>            components: clippy
>  
> -      - name: Build with ${{ matrix.toolchain }}
> +      - name: Build with ${{ matrix.toolchain }}, ${{ matrix.features }}
>          run: |
>            cargo +${{ matrix.toolchain }} build \
> -              --release --all-features --all-targets
> +              --release ${{ matrix.features }} --all-targets
>  
> -      - name: Test with ${{ matrix.toolchain }}
> +      - name: Test with ${{ matrix.toolchain }}, ${{ matrix.features }}
>          run: |
>            if [[ "${{ matrix.toolchain }}" == 'nightly' ]]; then
>                export RUSTDOCFLAGS='-Zsanitizer=address'
> @@ -47,7 +48,7 @@ jobs:
>            fi
>            cargo +${{ matrix.toolchain }} test \
>                --target=x86_64-unknown-linux-gnu \
> -              --all-features
> +              ${{ matrix.features }}
>  
>        - name: Lint with ${{ matrix.toolchain }}
>          run: |
> ```
> 
> I've not tested it, but if I didn't make any mistake it should add a workflow job to build and run the (available) tests without the `std` feature.

Thank you very much for proposing this change, It would have taken me a lot longer to figure out how the ci works and how to enable the `no_std` in the tests workflow. I realise that the last pipeline has failed but that was because I was missing the toolchain section in the `include`. Now I think it should work correctly, the test have failed because there was not toolchain specified in the last `no_std` test run

qmonnet:
The tests pass correctly this time :tada:, but we get warnings about `cargo-clippy`. I'm not sure why, maybe using `--no-default-features` disables something clippy-related. I'll take a look later.

SzymonKubica:
> The tests pass correctly this time 🎉, but we get warnings about `cargo-clippy`. I'm not sure why, maybe using `--no-default-features` disables something clippy-related. I'll take a look later.

I think the reason this started producing warnings was that running `cargo clippy` implicitly adds a `cargo-clippy` feature that isn't included in the `Cargo.toml`. It looks like `rbpf` uses a legacy way of specifying clippy lint attributes (possible explanation [here](https://blog.rust-lang.org/2024/02/28/Clippy-deprecating-feature-cargo-clippy.html)). Because of this, when running with --no-default-feature this `cargo-clippy` feature is missing and we get warnings. 

I also looked through the test logs and got rid of all other warnings inside of misc.rs and rbpf_plugin.rs. All of them were mostly unused imports or functions that weren't used when running in `no_std`.

After these changes all builds were passing and tests didn't introduce any new warnings. However,  I needed to squash the fix commits to make a single one that deals with all of the warnings. Could you please approve this workflow so that we can see that indeed everything got fixed? Thank you for your time, it's starting to get very exciting as we are getting closer having everything integrated :tada:

SzymonKubica:
Ok @qmonnet I think all checks were successful and after running tests with `--no-default-features` there were no warnings emitted :tada: :rocket: . Thank you for helping me enable tests for the `no_std` feature. Do let me know if anything else requires attention in the PR and I will adjust it.

qmonnet:
Thanks a lot! Sorry I didn't have the time to review last night, I'll do it soon.

SzymonKubica:
> Thanks a lot! Sorry I didn't have the time to review last night, I'll do it soon.

No worries at all, take your time. No need to rush, I realise it is quite a big change in terms of files affected (not that big functionality-wise). Thank you for your time and guidance with the PR.

src/lib.rs
```diff
@@ -59,6 +67,61 @@ mod interpreter;
...
+    pub use alloc::vec::Vec;
+    #[cfg(feature = "std")]
+    pub use std::vec::Vec;
+
+    #[cfg(not(feature = "std"))]
+    pub use alloc::string::{String, ToString};
+    #[cfg(feature = "std")]
+    pub use std::string::{String, ToString};
+
+
```
qmonnet, line 103: Nit: Double empty line, I think I've been trying to avoid these in the rest of the repo (although a few sneaked in)

src/no_std_error.rs
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+}
+
+/// The current version of `rbpf` only uses the [`Other`](ErrorKind::Other) variant
+/// from the [std::io::ErrorKind] enum. If a dependency on other variants were
+/// introduced in the future, this enum needs to be updated accordingly to maintain
+/// compatibility with the real `ErrorKind`. The reason all available variants
+/// aren't included in the first place is that [std::io::ErrorKind] exposes
+/// 40 variants, and not all of them are meaningful under `no_std`.
```
qmonnet, line 35: Thanks for documenting this! :+1:

tests/assembler.rs
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 // Copyright 2017 Rich Lane <lanerl@gmail.com>
 
+// These tests should only be run on std as they depend on the assembler.
+#![cfg(feature = "std")]
```
qmonnet, line 5: I'm confused here as well, can't we enable some of these tests now?

README.md
```diff
@@ -559,6 +560,26 @@ program.add(Source::Imm, Arch::X64).set_dst(1).set_imm(0x605).push()
...
+
+The `rbpf` crate has a Cargo feature named "std" that is enabled by default. To
+use `rbpf` in `no_std` environments this feature needs to be disabled. To do
+this, you need to modify your dependency on `rbpf` in Cargo.toml to disable the
+enabled-by-default features.
+```toml
+[dependencies]
+rbpf = { version = "1.0", default-features = false }
+```
+Note that when using this crate in `no_std` environments, the `assembler`,
```
qmonnet, line 575: Nit: Blank lines around the literal block, please

````suggestion
enabled-by-default features.

```toml
[dependencies]
rbpf = { version = "1.0", default-features = false }
```

Note that when using this crate in `no_std` environments, the `assembler`,
````

File: src/lib.rs [hunk 39]
```diff
@@ -43,11 +51,11 @@ extern crate cranelift_module;
...
 use byteorder::{ByteOrder, LittleEndian};
-use std::collections::HashMap;
-use std::io::{Error, ErrorKind};
-use std::u32;
+use crate::lib::*;
 
+#[cfg(feature = "std")]
 mod asm_parser;
+#[cfg(feature = "std")]
 pub mod assembler;
```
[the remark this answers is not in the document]
qmonnet: Ah but I see you re-enable it in one of the next commits, sorry.

qmonnet (changes requested):
Thanks a lot for your work!

I've got some comments, and I think I'm a bit confused as to when the assembler is supported without `std`. I probably need to take some time to play with your code and check why rbpf complains if I try to enable the tests; in the meantime, I commented at the locations where it is not clear to me.

Cargo.toml
```diff
@@ -44,11 +49,31 @@ hex = "0.4.3"
...
+[[example]]
+name = "uptime"
+required-features = ["std"]
+
+[[example]]
+name = "to_json"
+
+[[example]]
+name = "rbpf_plugin"
```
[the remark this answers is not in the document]
SzymonKubica: Yes, unfortunately we do. The way it works is that once you explicitly specify one example in `Cargo.toml` all other examples need to be specified as well. I remember seeing some warning when I didn't include these. I'm going to find it and put it here.

README.md
```diff
@@ -497,19 +504,21 @@ Conversely, a disassembler is also available to dump instruction names from
 bytecode in a human-friendly format.
 
 ```rust
-extern crate rbpf;
-use rbpf::disassembler::disassemble;
+#[cfg(feature = "std")] {
+    extern crate rbpf;
+    use rbpf::disassembler::disassemble;
```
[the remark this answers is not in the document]
SzymonKubica: Thank you for picking up on this one, yes I think this one should be good. I will confirm locally  and update the PR

src/asm_parser.rs
```diff
@@ -4,7 +4,9 @@
 // Rust-doc comments were left in the module, but it is no longer publicly exposed from the root
 // file of the crate. Do not expect to find those comments in the documentation of the crate.
 
-//! This module parses eBPF assembly language source code.
+//! This module parses eBPF assembly language source code. It depends on the
+//! `combine` crate, which depends on the standard library. Because of this
+//! it is only available when the `std` feature is enabled.
```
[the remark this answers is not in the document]
SzymonKubica: Yes I will amend this

src/lib.rs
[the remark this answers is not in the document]
SzymonKubica: Oh yes, sorry about this. I think it was originally disabled but then I tried harder to support it and it was probably included in one of the subsequent logical changes

src/lib.rs
SzymonKubica: Sure I will remove that, thanks for picking up on these!

Cargo.toml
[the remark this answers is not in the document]
SzymonKubica: This is the error message that pops up when examples aren't explicitly specified: 
```
warning: An explicit [[example]] section is specified in Cargo.toml which currently
disables Cargo from automatically inferring other example targets.
This inference behavior will change in the Rust 2018 edition and the following
files will be included as a example target:

* /home/szymon/Projects/ebpf-on-microcontrollers/vms/rbpf-for-microcontrollers/examples/to_json.rs
* /home/szymon/Projects/ebpf-on-microcontrollers/vms/rbpf-for-microcontrollers/examples/rbpf_plugin.rs

This is likely to break cargo build or cargo test as these files may not be
ready to be compiled as a example target today. You can future-proof yourself
and disable this warning by adding `autoexamples = false` to your [package]
section. You may also move the files to a location where Cargo would not
automatically infer them to be a target, such as in subfolders.

```

src/assembler.rs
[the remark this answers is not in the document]
SzymonKubica: Yes this got out of date, will update now

src/cranelift.rs
```diff
@@ -28,6 +21,7 @@ use crate::ebpf::{
 };
 
 use super::Error;
+use crate::lib::*;
```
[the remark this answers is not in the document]
SzymonKubica: Totally agree, that way all crate imports are in one place.

src/disassembler.rs
```diff
@@ -383,6 +386,7 @@ pub fn to_insn_vec(prog: &[u8]) -> Vec<HLInsn> {
 /// neg64 r2
 /// exit
 /// ```
+#[cfg(feature = "std")]
 pub fn disassemble(prog: &[u8]) {
```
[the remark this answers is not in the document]
SzymonKubica: I have updated it so that it calls log::info! when under no_std. I think it basically replicates the behaviour without disabling it. The idea is that in `no_std` one can define a custom logger that calls into the OS-specific printing functionality that can then print to a log file or the serial port like so:
```
impl Log for RiotLogger {
    fn enabled(&self, metadata: &Metadata<'_>) -> bool {
        metadata.level() <= self.level
    }

    fn log(&self, record: &Record<'_>) {
        if self.enabled(record.metadata()) {
            println!(
                "{}:{} -- {}",
                record.level(),
                record.target(),
                record.args()
            );
        }
    }

    fn flush(&self) {}
}
```
The example above is the logger that I used for RIOT applications.

tests/assembler.rs
SzymonKubica: I think those were reenabled in one of the subsequent commits. I decided treat enabling no_std in the assembler and disassembler (and enabling their corresponding tests) as a separate feature as it required changing the way we use `combine` library. I can squash the two commits together so that it is no longer confusing

SzymonKubica:
> Thanks a lot for your work!
> 
> I've got some comments, and I think I'm a bit confused as to when the assembler is supported without `std`. I probably need to take some time to play with your code and check why rbpf complains if I try to enable the tests; in the meantime, I commented at the locations where it is not clear to me.
> 
> The required changes are mostly minor things, we're getting closer!
> 
> Regarding the formatting of your patches, I'd recommend the following:
> 
> * Squash the fixes from the last commit into the relevant earlier commits
> * Rework your commit descriptions, in particular for the second commit. What's interesting in this description is not so much to describe the changes (and not the history of the different changes and fixes incorporated during the review); instead, it should provide the motivation for the change, and explain what the commit addresses, what difficulties arose and how the commit works around them, etc. Everything related to the context of the change.

Ok thank you for the review and your suggestions. I will rework the commit messages and squash the subsequent fixes.
I propose the following logical changes:
- add the `std` feature and updating the windows cfg flags
- add lib module and conditional imports (the **Configure combine crate and asm_parser to work under no_std.** commit will be squashed into this one)
- update readme with feature description
- add workflow job

The last commit fixing pipeline warnings will be split and squashed into the above commits where the changes belongj appropriately

tests/assembler.rs
SzymonKubica: Oh I doublechecked, sorry about this. Those can indeed be enabled without issues

SzymonKubica:
Ok I think everything has been cleaned up now. Thank you very much for all of your feedback and time spent reviewing my changes. Now all imports / dependency changes are in the second commit. I managed to find some additional assembler tests that could be reenabled (tests/misc.rs). I think it helps a lot to see all changes in a single logical commit

SzymonKubica:
@qmonnet could you please approve the CI workflow run? Don't worry about reviewing the changes today, no need to rush. I just wanted to make sure that rearranging the changes didn't produce any new pipeline warnings. Enjoy your weekend and feel free to look into the PR when you are free.

src/disassembler.rs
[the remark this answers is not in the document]
qmonnet: > The example above is the logger that I used for RIOT applications.

Just to double-check and be 100% clear here, this is some code that _you_ contributed to RIOT, not something in RIOT from which you took inspiration, right? I'm asking because I see RIOT is LGPL, so we can't reproduce its code in rbpf if you weren't the one to contribute it in the first place (although reading again, you mention RIOT _applications_ which is maybe something different).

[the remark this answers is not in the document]
SzymonKubica: Yes that's right, this is just some custom application-specific code that I have in an app built on top of riot. Additionally, this snippet of code was just to illustrate to you how one might go about implementing the logger behaviour in their system. It's not actually contributed anywhere in rbpf. There is nothing in this PR that was taken / inspired by RIOT, I was just trying to motivate why using the log crate is a good idea here to allow users to specify where things should get logged to

README.md
```diff
@@ -555,6 +556,28 @@ program.add(Source::Imm, Arch::X64).set_dst(1).set_imm(0x605).push()
...
+
+```toml
+[dependencies]
+rbpf = { version = "1.0", default-features = false }
+```
+
+Note that when using this crate in `no_std` environments, the `assembler`,
+`asm_parser` and `jit` modules aren't available. This is because the assembler
+depends on the `combine` crate providing parser combinators. This crate doesn't
+provide `no_std` support for the features that `rbpf` depends on.
```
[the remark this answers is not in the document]
SzymonKubica: No worries at all, I realise that maintaining an open source project alongside your main job is quite challenging. Thank you for all of your feedback in getting this PR into an acceptable state. I will quickly amend that commit to reflect the current state of the system. Thank you for picking up on it, I really appreciate the level of detail that you pay attention to.

src/disassembler.rs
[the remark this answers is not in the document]
qmonnet: OK perfect, thank you for confirming :slightly_smiling_face:

SzymonKubica:
> Apologies once again for the delay. I do realise this is not a great way to keep you motivated for this PR, and I'm sorry for the poor contribution experience in terms of delays. Thanks a lot for bearing with me.
> 
> And thanks for this work, this is in a very nice shape now! I think there's just a small fix to do in the README, but everything else looks good as far as I can tell. Well done, and nice work on the commit description for your second commit! ❤️

Ok @qmonnet, thank you for pointing out the issue it the README, it is fixed now. Thank you for picking up on it.

qmonnet (approved):
Awesome, thanks a lot for this work!

SzymonKubica:
Perfect, thank you for approving the PR :rocket: 
I really appreciate your help along the way. It was a great learning experience and I think it is going to be very useful for my future open-source contributions. Thanks for your time and being open to this contribution.

SzymonKubica:
> Congrats on getting your µBPF paper into the ACM workshop! 🎉

Thank you!

PR #107, opened 2024-04-19
README.md: Update link referencing JIT

60ke:
I noticed that the  link has expired. Perhaps this link is correct, but I'm not sure


qmonnet (changes requested):
Good catch, thank you! The link looks correct. I understand the author has changed name, so we might just as well use her new one I suppose? Would you mind changing “Jonathan” to “Sophia” on the next line, please?

60ke:
> Good catch, thank you! The link looks correct. I understand the author has changed name, so we might just as well use her new one I suppose? Would you mind changing “Jonathan” to “Sophia” on the next line, please?
yes, done it

qmonnet (approved):

qmonnet:
I added the author's last name back, and squashed the commits. Thanks!

PR #108, opened 2024-05-01
src/ebpf.rs: Increase instruction limit to 1 million

seanyoung:
The linux kernel allows programs of 1 million instructions, see

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=c04c0d2b968ac45d6ef020316808ef6c82325a82

I'm hitting the 4096 limit in some corner cases with https://github.com/seanyoung/cir, if the infrared protocol is very complex (e.g. XMP). rbpf is used for testing the generated IR decoders in the test framework.


qmonnet (changes requested):
Thanks!

> The linux kernel allows programs of 1 million instructions, see
> 
> https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=c04c0d2b968ac45d6ef020316808ef6c82325a82

Not really, [this is a limit for the program complexity](https://qmonnet.github.io/whirl-offload/2023/11/05/dyk-prog-size-limit/) which means that in practice, you get either less than this, or a very boring program with no branch.

> I'm hitting the 4096 limit in some corner cases with https://github.com/seanyoung/cir, if the infrared protocol is very complex (e.g. XMP). rbpf is used for testing the generated IR decoders in the test framework.

OK, I'm not using the project much myself these days, but if it's helpful to you to raise the limit, I'm fine with that.

Please update test `test_verifier_err_too_many_instructions` in `tests/ubpf_verifier.rs`, though.

qmonnet (approved):
Thank you!

PR #109, opened 2024-05-25
Fix mislabelled mul64 opcodes

deanmlittle:
mul64 opcodes were labelled as div64 in the comments with an operation equivalent `/=`. They are now accurately labelled as mul64 with an operation equivalent to `*=`


qmonnet (approved):
Thank you! Could you please [sign off](https://wiki.linuxfoundation.org/dco) your commit?

PR #114, opened 2024-09-27
Reduce usage of libc

notgull:
I am attempting to reduce the dependence on libc in my indirect
dependencies. While looking at this crate I noticed some low-hanging
fruit that can be very easily replaced with libc.

- The rand() function can be re-implemented as a simple Wyrand seeded by
  the thread ID and the current time.
- libc::memcpy is used at a point where it can be replaced with
  ptr::copy_nonoverlapping.
- libc is used for memory allocation at a point where std::alloc can be
  used instead. So I use this.

I'm not too familiar with this codebase, so if any of these replacements
are incorrect, please let me know.

The only leftover libc call is mprotect().

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -16,9 +16,6 @@
 //! value. Hence some helpers have unused arguments, or return a 0 value in all cases, in order to
 //! respect this convention.
 
-#[cfg(feature = "std")]
-extern crate libc;
-
 use crate::lib::*;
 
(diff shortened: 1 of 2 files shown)


qmonnet (approved):
Apologies for the delay.

PR looks good to me, I've got nothing particular against libc but I'm not against getting rid of a dependency (or at least getting closer to it). Thanks for this work!

PR #122, opened 2025-04-01
feat: Add bpf to bpf call support

Godones:
Like ubpf, we can check the correctness of bpf calls and add support for bpf to bpf calls.

### DIFF
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -387,7 +387,7 @@ pub fn execute_program(
                         insn_ptr += insn.imm as usize;
                     }
                     _ => {
-                        Err(Error::new(ErrorKind::Other, format!("Error: invalid call to function #{:?} (insn #{insn_ptr:?})", insn.imm)))?;
+                        Err(Error::new(ErrorKind::Other, format!("Error: unexpected call type #{} (insn #{})", _src, insn_ptr-1)))?;
                     }
                 }
             }
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -6,7 +6,7 @@
 //      (Translation to Rust, MetaBuff/multiple classes addition, hashmaps for helpers)
 
 use crate::ebpf;
-use crate::ebpf::RBPF_MAX_CALL_DEPTH;
+use crate::ebpf::MAX_CALL_DEPTH;
 use crate::lib::*;
 use crate::stack::{StackFrame, StackUsage};
 
@@ -56,14 +56,15 @@ pub fn execute_program(
     const U32MAX: u64 = u32::MAX as u64;
     const SHIFT_MASK_64: u64 = 0x3f;
 
-    let (prog,stack_usage) = match prog_ {
+    let (prog, stack_usage) = match prog_ {
         Some(prog) => (prog, stack_usage.unwrap()),
         None => Err(Error::new(ErrorKind::Other,
                     "Error: No program set, call prog_set() to load one"))?,
     };
     let stack = vec![0u8;ebpf::STACK_SIZE];
-    let mut stacks = [StackFrame::new();RBPF_MAX_CALL_DEPTH];
+    let mut stacks = [StackFrame::new();MAX_CALL_DEPTH];
     let mut stack_frame_idx = 0;
+
     // R1 points to beginning of memory area, R10 to stack
     let mut reg: [u64;11] = [
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, stack.as_ptr() as u64 + stack.len() as u64
@@ -86,7 +87,7 @@ pub fn execute_program(
     let mut insn_ptr:usize = 0;
     while insn_ptr * ebpf::INSN_SIZE < prog.len() {
         let insn = ebpf::get_insn(prog, insn_ptr);
-        if stack_frame_idx < RBPF_MAX_CALL_DEPTH{
+        if stack_frame_idx < MAX_CALL_DEPTH{
             if let Some(usage) = stack_usage.stack_usage_for_local_func(insn_ptr) {
                 stacks[stack_frame_idx].set_stack_usage(usage);
             }
@@ -375,19 +376,22 @@ pub fn execute_program(
                             Err(Error::new(ErrorKind::Other, format!("Error: unknown helper function (id: {:#x})", insn.imm as u32)))?;
                         }
                     }
-                    // BPF To BPF call
+                    // eBPF to eBPF call
                     1 => {
-                        if stack_frame_idx >= RBPF_MAX_CALL_DEPTH {
-                            Err(Error::new(ErrorKind::Other, format!("Error: too many nested calls (max: {RBPF_MAX_CALL_DEPTH})")))?;
+                        if stack_frame_idx >= MAX_CALL_DEPTH {
+                            Err(Error::new(ErrorKind::Other, format!("Error: too many nested calls (max: {MAX_CALL_DEPTH})")))?;
                         }
                         stacks[stack_frame_idx].save_registers(&reg[6..=9]);
                         stacks[stack_frame_idx].save_return_address(insn_ptr);
+                        // Why we don't need to check the stack usage here?
+                        // When the stack is exhausted, if there are instructions in the new function
+                        // that read or write to the stack, check_mem_load or check_mem_store will return an error.
                         reg[10] -= stacks[stack_frame_idx].get_stack_usage().stack_usage() as u64;
                         stack_frame_idx += 1;
                         insn_ptr += insn.imm as usize;
                     }
                     _ => {
-                        Err(Error::new(ErrorKind::Other, format!("Error: unexpected call type #{} (insn #{})", _src, insn_ptr-1)))?;
+                        Err(Error::new(ErrorKind::Other, format!("Error: unsupported call type #{} (insn #{})", _src, insn_ptr-1)))?;
                     }
                 }
             }
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -6,7 +6,9 @@
 //      (Translation to Rust, MetaBuff/multiple classes addition, hashmaps for helpers)
 
 use crate::ebpf;
+use crate::ebpf::RBPF_MAX_CALL_DEPTH;
 use crate::lib::*;
+use crate::stack::{StackFrame, StackUsage};
 
 #[allow(clippy::too_many_arguments)]
 fn check_mem(
@@ -45,6 +47,7 @@ fn check_mem(
 
 pub fn execute_program(
     prog_: Option<&[u8]>,
+    stack_usage: Option<&StackUsage>,
     mem: &[u8],
     mbuff: &[u8],
     helpers: &HashMap<u32, ebpf::Helper>,
@@ -53,13 +56,14 @@ pub fn execute_program(
     const U32MAX: u64 = u32::MAX as u64;
     const SHIFT_MASK_64: u64 = 0x3f;
 
-    let prog = match prog_ {
-        Some(prog) => prog,
+    let (prog,stack_usage) = match prog_ {
+        Some(prog) => (prog, stack_usage.unwrap()),
         None => Err(Error::new(ErrorKind::Other,
                     "Error: No program set, call prog_set() to load one"))?,
     };
     let stack = vec![0u8;ebpf::STACK_SIZE];
-
+    let mut stacks = [StackFrame::new();RBPF_MAX_CALL_DEPTH];
+    let mut stack_frame_idx = 0;
     // R1 points to beginning of memory area, R10 to stack
     let mut reg: [u64;11] = [
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, stack.as_ptr() as u64 + stack.len() as u64
@@ -82,6 +86,11 @@ pub fn execute_program(
     let mut insn_ptr:usize = 0;
     while insn_ptr * ebpf::INSN_SIZE < prog.len() {
         let insn = ebpf::get_insn(prog, insn_ptr);
+        if stack_frame_idx < RBPF_MAX_CALL_DEPTH{
+            if let Some(usage) = stack_usage.stack_usage_for_local_func(insn_ptr) {
+                stacks[stack_frame_idx].set_stack_usage(usage);
+            }
+        }
         insn_ptr += 1;
         let _dst = insn.dst as usize;
         let _src = insn.src as usize;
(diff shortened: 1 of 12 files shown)


qmonnet (commented):
Thanks! Looks good, although I haven't looked into the details yet.

I'd really like to have support in the JIT before merging function calls support though, or we'll lose feature parity.

Godones:
Ok, maybe I can try to add JIT related code although I am not very familiar with JIT.

qmonnet:
It's not trivial, you'll have to look into some assembly code - but it would be awesome if you could do that, thanks for looking!

[uBPF supports it](https://github.com/iovisor/ubpf/commit/2e3b03952333dfd9c2ccff20e05a33ad1bbaa407), you're welcome to base your code on theirs (the JIT in rbpf is largely adapted from uBPF already); no kernel code though (GPL).

Godones:
Thank you, it doesn't look very difficult. I should be able to solve it.

Godones:
Hi, I've added support for function calls to jit and written two simple tests.

Godones:
Hi, do you have time to take a look at this PR. I would like to add some features after this PR is received.

src/stack.rs
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+    /// Validate the stack usage of a program
+    ///
+    /// This function checks the stack usage of a program and returns a `StackUsage` object
+    /// containing the stack usage for each local function in the program.
+    ///
+    /// # Returns
+    /// - `Ok(StackUsage)` if the stack usage is valid
+    /// - `Err(Error)` if the stack usage is invalid
```
qmonnet, line 92: It would be good to explain what makes stack usage to be invalid. As far as I understand, with your code it is valid if and only if the size if a multiple of 16, right?

src/stack.rs
```diff
@@ -0,0 +1,140 @@
...
+        &mut self,
+        prog: &[u8],
+        pc: usize,
+    ) -> Result<StackUsageType, Error> {
+        let mut ty = StackUsageType::Default;
+        if let Some(calculator) = self.calculator {
+            let size = calculator(prog, pc, self.data.as_mut().unwrap());
+            ty = StackUsageType::Custom(size);
+        }
+        if ty.stack_usage() % 16 > 0 {
```
qmonnet, line 119: Why 16?

src/stack.rs
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /// Calculate the stack usage for a local function
+    fn calculate_stack_usage_for_local_func(
+        &mut self,
+        prog: &[u8],
+        pc: usize,
+    ) -> Result<StackUsageType, Error> {
+        let mut ty = StackUsageType::Default;
+        if let Some(calculator) = self.calculator {
```
qmonnet, line 115: Or let's return `Ok(StackUsageType::Default)` immediately if we don't have `self.calculator`, given that we know we'll have a good size anyway in that case?

src/lib.rs
```diff
@@ -116,6 +118,9 @@ pub type Verifier = fn(prog: &[u8]) -> Result<(), Error>;
 /// eBPF helper function.
 pub type Helper = fn(u64, u64, u64, u64, u64) -> u64;
 
+/// eBPF stack usage calculator function.
+pub type StackUsageCalculator = fn(prog:&[u8], pc:usize, data:&mut dyn Any) -> u16;
```
qmonnet, line 127: What's the use case for this? It would be good to mention it in the commit description. It's not obvious to me.

src/lib.rs
```diff
@@ -66,7 +68,7 @@ pub mod lib {
     pub use self::core::mem;
     pub use self::core::mem::ManuallyDrop;
     pub use self::core::ptr;
-
+    pub use core::any::Any;
     pub use self::core::f64;
```
qmonnet, line 72: `self::core::any::Any` for consistency, and let's sort this block alphabetically while at it?

tests/ubpf_vm.rs
```diff
@@ -2227,6 +2228,19 @@ fn test_vm_stxw() {
...
+    program.mov(Source::Imm, Arch::X64).set_dst(0x1).set_imm(0xff).push()
+    .call().set_src(1).set_imm(1).push()
+    .mov(Source::Imm, Arch::X64).set_dst(0x2).set_imm(0x1).push()
+    .mov(Source::Imm, Arch::X64).set_dst(0).set_imm(0xf).push()
+    .exit().push();
+    let prog = program.into_bytes();
+    let vm = rbpf::EbpfVmNoData::new(Some(&prog)).unwrap();
+    assert_eq!(vm.execute_program().unwrap(), 0xf);
+}
+
```
qmonnet: The assembly syntax would be more intuitive to read. Did you fall back on the chained API because the assembler doesn't support function calls?

tests/ubpf_vm.rs
```diff
@@ -2227,6 +2228,19 @@ fn test_vm_stxw() {
     assert_eq!(vm.execute_program(mem).unwrap(), 0x44332211);
 }
 
+#[test]
+fn test_bpf_to_bpf_call(){
```
qmonnet, line 2232: Is this test inspired from uBPF's tests? If not, can you please move it to tests/misc.rs instead?

tests/ubpf_vm.rs
```diff
@@ -2230,15 +2230,35 @@ fn test_vm_stxw() {
...
+    let mut real_code = Vec::new();
+    code.iter_mut().for_each(|insn| {
+        if insn.opc == CALL {
+            insn.src = 0x1;
+        }
+        real_code.extend_from_slice(&insn.to_array());
+    });
+    let vm = rbpf::EbpfVmNoData::new(Some(&real_code)).unwrap();
+    let vm_res= vm.execute_program().unwrap();    
+    assert_eq!(vm_res, 0x10);
```
qmonnet, line 2261: OK, this change is not related to the JIT and should go into the first commit.

tests/ubpf_vm.rs
```diff
@@ -2230,15 +2230,35 @@ fn test_vm_stxw() {
...
+    )
+    .unwrap();
+    let mut code = to_insn_vec(&test_code);
+    let mut real_code = Vec::new();
+    code.iter_mut().for_each(|insn| {
+        if insn.opc == CALL {
+            insn.src = 0x1;
+        }
+        real_code.extend_from_slice(&insn.to_array());
+    });
```
qmonnet, line 2258: Yikes, we'll have to complete the assembler (and disassembler). We can do as a follow-up if you don't want to tackle it here, though.

Ideally we should preserve the current syntax for helper function calls, but find something to allow BPF-to-BPF calls, too. I don't remember what syntax the kernel disassembler or llvm use for function calls, would be nice to not deviate too much :thinking:

src/jit.rs
```diff
@@ -461,6 +461,20 @@ impl JitCompiler {
         }
     }
 
+    fn emit_local_call(&mut self, mem: &mut JitMemory, target_pc: isize) {
+        self.emit_push(mem, map_register(6));
+        self.emit_push(mem, map_register(7));
+        self.emit_push(mem, map_register(8));
+        self.emit_push(mem, map_register(9));
+        // e8 is the opcode for a CALL
```
qmonnet, line 469: Nit:

```suggestion
        // 0xe8 is the opcode for a CALL
```

src/jit.rs
```diff
@@ -516,6 +530,10 @@ impl JitCompiler {
         // Allocate stack space
         self.emit_alu64_imm32(mem, 0x81, 5, RSP, ebpf::STACK_SIZE as i32);
 
+        self.emit1(mem, 0xe8);
+        self.emit4(mem, 5);
+        self.emit_jmp(mem, TARGET_PC_EXIT);
```
qmonnet, line 535: Can you please add a comment to explain what this does? It creates a call to the “function” that starts right after the jump to `TARGET_PC_EXIT` so that the final `EXIT` instruction comes back here, right?

tests/ubpf_jit_x86_64.rs
```diff
@@ -2391,3 +2391,19 @@ fn test_bpf_to_bpf_call(){
...
+        0x95, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    ];
+    let mut vm = rbpf::EbpfVmNoData::new(None).unwrap();
+    vm.set_verifier(|_|{
+        Ok(())
+    }).unwrap();
+    vm.set_program(prog).unwrap();
+    vm.jit_compile().unwrap();
+    unsafe { vm.execute_program_jit().unwrap() };
+}
```
qmonnet, line 2409: Missing end-of-line character at the end of this file (same also applies for the other two test files).

tests/ubpf_vm.rs
```diff
@@ -2420,3 +2420,18 @@ fn test_vm_tcp_sack_nomatch() {
     let vm = rbpf::EbpfVmRaw::new(Some(&prog)).unwrap();
     assert_eq!(vm.execute_program(mem.as_mut_slice()).unwrap(), 0x0);
 }
+
+#[test]
+#[should_panic(expected = "Error: unexpected call type #2 (insn #0)")]
+fn test_other_type_call(){
```
qmonnet, line 2426: Can we also have tests where we mess with the stack size, please? Invalid stack validation, or running out of stack space, for example.

qmonnet (changes requested):
Great work, thank you!

I've got a number of comments, don't worry - most of them are minor. A few are more important:

- I think we miss a check on remaining stack size when doing a function call, in the interpreter
- Would be nice to get a few more tests, to check what happens when we mess up with stack size
- We miss the update for the assembler, disassembler, crane-based JIT compiler (but that's OK if they're not in this PR)
- I'm not sure I understand the use case for the custom stack size validator, would you mind explaining please?

As a side note, you mentioned you had other features in mind, what are they? And if I may ask, what's your use case for rbpf?

src/interpreter.rs
```diff
@@ -356,13 +365,43 @@ pub fn execute_program(
...
+                        }
+                    }
+                    // BPF To BPF call
+                    1 => {
+                        if stack_frame_idx >= RBPF_MAX_CALL_DEPTH {
+                            Err(Error::new(ErrorKind::Other, format!("Error: too many nested calls (max: {RBPF_MAX_CALL_DEPTH})")))?;
+                        }
+                        stacks[stack_frame_idx].save_registers(&reg[6..=9]);
+                        stacks[stack_frame_idx].save_return_address(insn_ptr);
+                        reg[10] -= stacks[stack_frame_idx].get_stack_usage().stack_usage() as u64;
```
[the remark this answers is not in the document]
Godones: `check_mem_load` and `check_mem_store` will check if there are instructions accessing the stack later.

src/stack.rs
Godones: yes

src/stack.rs
Godones: This check references ubpf, although I'm not sure why.
https://github.com/iovisor/ubpf/blob/8e5b6127405b7383667976c3dd1c09e82dfa86a1/vm/ubpf_vm.c#L1926

src/lib.rs
Godones: A simple example is: when executing multiple simple func calls, since these functions usually do not use the stack, in theory, the stack space will not be exhausted. However, our current stack calculation algorithm is static and cannot handle this situation, so we allow users to customize their stack calculation method.

src/jit.rs
Godones: Yes, I'll add a comment.

src/verifier.rs
```diff
@@ -248,7 +248,21 @@ pub fn check(prog: &[u8]) -> Result<(), Error> {
             ebpf::JSLE_IMM32 => { check_jmp_offset(prog, insn_ptr)?; },
             ebpf::JSLE_REG32 => { check_jmp_offset(prog, insn_ptr)?; },
 
-            ebpf::CALL       => {},
+            ebpf::CALL       => {
+                let src = insn.src;
+                match src {
+                    0 => {
+                        if insn.imm < 0 { reject(format!("invalid call to function #{:?} (insn #{insn_ptr:?})", insn.imm))?; }
```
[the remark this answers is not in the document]
Godones: I think it's generally a mistake to treat negative numbers as unsigned. I think we can check for this in the verifier.

tests/ubpf_vm.rs
Godones: Yes, I added the necessary support in the assembler and disassembler.

Godones:
The new feature I would like to add is to allow jit to be used in a no_std environment. In this way, we can use rbpf in rust-implemented os.

src/verifier.rs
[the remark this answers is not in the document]
qmonnet: I don't think it matters much here, as they are just IDs for helper functions. If a user wanted to use the full range of available ID, whether they translate as unsigned or a signed negative number, it doesn't really matter here - I see no reason to block that (do you?).

src/interpreter.rs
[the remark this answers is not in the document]
qmonnet: Hmm so basically you allow to keep going with the function call, in case the callee doesn't use the stack, and only raise an error when attempting to read or write out of the stack? Alright, that should be OK - but please let's add a comment here to explain why it's OK to skip the check.

src/stack.rs
qmonnet: Yep I noticed, but it would be nice to check whether it's relevant in our case. I asked [here](https://github.com/iovisor/ubpf/pull/271#issuecomment-2799037009).

src/lib.rs
qmonnet: OK thanks, that's what I thought.
- Do you actually plan on using this feature? If not, and given that nobody's been using the function calls with rbpf this far, I'm wondering if we should support customising this at all, for now.
- If you do intend to use it, please provide more details (what you commented here, for example) in the commit description, so that we can understand the reason for the changes.

tests/misc.rs
```diff
@@ -602,3 +602,126 @@ fn test_verifier_fail() {
...
+        add64 r0, r1
+        exit
+        ").unwrap();
+    let vm = rbpf::EbpfVmNoData::new(Some(&test_code)).unwrap();
+    let vm_res= vm.execute_program().unwrap();
+    assert_eq!(vm_res, 0x10);
+}
+
+#[test]
+fn test_vm_jit_bpf_to_bpf_call(){
```
qmonnet, line 630: Thanks for moving these! Contrarily to the test file with the JIT tests inherited from uBPF, this file doesn't have a toplevel `#![cfg(all(not(windows), feature = "std"))]
`, which means we must manually exclude tests using the JIT from `windows` and `no_std`. This should fix the Windows build.


```suggestion
#[cfg(all(not(windows), feature = "std"))]
#[test]
fn test_vm_jit_bpf_to_bpf_call(){
```

tests/misc.rs
```diff
@@ -602,3 +602,126 @@ fn test_verifier_fail() {
...
+    vm.set_verifier(|_|{
+        Ok(())
+    }).unwrap();
+    vm.set_program(prog).unwrap();
+    vm.execute_program().unwrap();
+}
+
+#[test]
+#[should_panic(expected = "[JIT] Error: unexpected call type #2 (insn #0)")]
+fn test_vm_jit_other_type_call(){
```
qmonnet, line 680:
```suggestion
#[cfg(all(not(windows), feature = "std"))]
#[test]
#[should_panic(expected = "[JIT] Error: unexpected call type #2 (insn #0)")]
fn test_vm_jit_other_type_call(){
```

src/verifier.rs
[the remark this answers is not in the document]
Godones: Yes, your opinion is correct, I will revert this modification.

src/lib.rs
Godones: I added a detailed description in my first commit.

qmonnet (commented):
Looks in good shape, thanks! Could I ask you to move the fixes from commits 4 and 5 to the relevant, previous commits, so that we end up with only 4 commits, and commit 4 just contains the assembler/disassembler bits (maybe adjust the commit title as well), please?

Let's also give uBPF folks a couple days to reply. This 16-bytes alignment is bothering me, in particular because they also make sure to preserve it across function calls, which is not the case in this PR.

src/lib.rs
qmonnet: It doesn't reply to the first part of the question though :stuck_out_tongue:

qmonnet: Thanks!

Sorry, I didn't mean to question the usefulness of BPF-to-BPF calls, I'm super happy you've been working on this. I just meant the custom stack usage validator.

Godones: I think a custom stack calculator is useful. Of course, if we had an accurate way to calculate the stack used by a function, we wouldn't need it.

qmonnet: Ok. Yes, calculating the stack, and properly validating the program by the way, would be nice.

src/stack.rs
Godones: Hi, I noticed that the developers of ubpf seem to have not answered this question yet. Maybe I can remove the 16-byte alignment requirement first and add it later if necessary. This way we can continue to push new features.

qmonnet: I can chase up with them, or we can go without the 16-byte restriction for now indeed, whichever you prefer. Apologies for the delay.

qmonnet: Apologies, it seems the contributor who introduced the 16-byte restriction is busier than I expected.

If you're still interested - I know this sort of delays are a killer for motivation :disappointed: - please let's drop the alignment requirement for now and go ahead with the rest of the PR.

Godones: Let's keep moving forward.

qmonnet (approved):
Looks good, thank you!

Just one question:

> I think a custom stack calculator is useful. Of course, if we had an accurate way to calculate the stack used by a function, we wouldn't need it.

Do you actually _use_ the custom stack calculator? If not, and especially now that we removed the alignment requirement and don't do any checks, I consider dropping it until someone comes up with a use case for it (I can do the change after we've merged the PR in its current state though).

Godones:
Sorry, I don't have an example right now, but I think one will come soon. Let's keep it for now, shall we?

PR #125, opened 2025-05-11
feat: support JIT in no_std mode

Godones:
In order to allow the use of jit in no_std mode, it is necessary to let the user manually pass in the executable memory area.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,7 +46,7 @@ pub mod ebpf;
 pub mod helpers;
 pub mod insn_builder;
 mod interpreter;
-#[cfg(all(not(windows), feature = "std"))]
+#[cfg(not(windows))]
 mod jit;
 #[cfg(not(feature = "std"))]
 mod no_std_error;
@@ -79,6 +79,8 @@ pub mod lib {
     #[cfg(not(feature = "std"))]
     pub use alloc::vec::Vec;
     #[cfg(feature = "std")]
+    pub use std::vec;
+    #[cfg(feature = "std")]
     pub use std::vec::Vec;
 
     #[cfg(not(feature = "std"))]
@@ -108,6 +110,8 @@ pub mod lib {
 
     #[cfg(not(feature = "std"))]
     pub use alloc::format;
+    #[cfg(feature = "std")]
+    pub use std::format;
 }
 
 /// eBPF verification function that returns an error if the program does not meet its requirements.
@@ -171,7 +175,7 @@ struct MetaBuff {
 pub struct EbpfVmMbuff<'a> {
     prog: Option<&'a [u8]>,
     verifier: Verifier,
-    #[cfg(all(not(windows), feature = "std"))]
+    #[cfg(not(windows))]
     jit: Option<jit::JitMemory<'a>>,
     #[cfg(feature = "cranelift")]
     cranelift_prog: Option<cranelift::CraneliftProgram>,
@@ -209,7 +213,7 @@ impl<'a> EbpfVmMbuff<'a> {
         Ok(EbpfVmMbuff {
             prog,
             verifier: verifier::check,
-            #[cfg(all(not(windows), feature = "std"))]
+            #[cfg(not(windows))]
             jit: None,
             #[cfg(feature = "cranelift")]
             cranelift_prog: None,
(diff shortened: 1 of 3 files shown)


qmonnet:
[This is just to say that your PR is still on my radar, sorry for the wait]

src/lib.rs
```diff
@@ -371,6 +371,20 @@ impl<'a> EbpfVmMbuff<'a> {
...
+    ///
+    /// # Warning
+    /// This function will replace any previously registered helpers with the new set.
+    pub fn register_helper_set(
+        &mut self,
+        helpers: HashMap<u32, ebpf::Helper>,
+    ) -> Result<(), Error> {
+        self.helpers = helpers;
+        Ok(())
+    }
```
qmonnet, line 388: I'm not a fan of this, I don't really see the point and it exposes the internal structure containing the helpers (we could always change it internally, but then passing a HashMap to the function to convert it to something else would make no sense).

It should be easy enough and not add too much overhead to loop on your map and call `register_helper()` every time, shouldn't it?

README.md
```diff
@@ -210,7 +210,7 @@ useful for programs that should be compatible with the Linux kernel and
 therefore must use specific helper numbers.
 
 ```rust,ignore
-pub fn register_allowed_memory(&mut self,, addr: &[u64]) -> ()
+pub fn register_allowed_memory(&mut self,, addrs_range: Range<u64>) -> ()
```
qmonnet, line 213: Looks like the double comma is a typo, can you please fix it?

src/lib.rs
```diff
@@ -71,6 +71,7 @@ pub mod lib {
     pub use self::core::mem;
     pub use self::core::mem::ManuallyDrop;
     pub use self::core::ptr;
+    pub use hashbrown::{HashMap,HashSet};
```
qmonnet, line 74: Nit

```suggestion
    pub use hashbrown::{HashMap, HashSet};
```

qmonnet (changes requested):
Thank you! It's great to gain support for the JIT without `std`. I'm in favour of taking this work (again, apologies for the delay!), but I've got some concerns with the current version.

- I'm not convinced we need the second commit (helpers batch registration), it exposes the internal structure we use to hold the helper functions, for little gain in my opinion.

- I don't fully understand the motivation for commit 3. What are the ranges trying to solve, here?
  > Since some maps may store a lot of data, representing the address with a
  > single byte may waste a lot of space. Using range can simplify this
  > problem.

  We don't support maps yet. And what do you mean by “a single byte” exactly?

- Any chance you could reorder the commits to avoid breaking the tests between commits 3 and 4? By folding the last commit into the previous one, or swapping them if it makes sense. I don't like broken tests between commits, it's a pain when we need to bisect to find a bug.

- We need to update the tests, too, so we can be sure that the JIT behaves as expected. The good news is, it's easy to remove the `feature = "std"` in `tests/ubpf_jit_x86_64.rs` and `test/misc.rs`. Bad news: we need to add macros and use the alternate version of the `jit_compile()` function everywhere.

Regarding this last point: Would it make sense to maybe change your API? Instead of having a `jit_compile()` function with different arguments when `std` is disabled, how about adding a separate `set_jit_exec_memory()` to add the memory area as an attribute to the VM, and use it in `jit_compile()` without having to pass it at compilation time? Makes one more function call when we don't have `std`, but it avoids making different versions of the function, and look slightly cleaner to me? We'd still have to update all the JIT tests though.

src/lib.rs
Godones: Yes, it is better to keep it simple

Godones:
For [commit 2](https://github.com/qmonnet/rbpf/pull/125/commits/0b79cac2856c413a40fdc25909639ea946f4e275), suppose we have an array data type as follows: 
```
struct Array{
    data: [u8;512]
}
```

The size of this array is 512 bytes. If we follow the original method, we need to store 512 u64 types, which requires 512\*8 space. Using Range<usize>, only 8\*2 space is required.

Godones:
It is a bit difficult to test the JIT in no_std environment, because executable memory allocation is usually kernel-specific. I don't think the lack of this test will affect the use of the JIT in no_std mode.

qmonnet:
> The size of this array is 512 bytes. If we follow the original method, we need to store 512 u64 types, which requires 512*8 space. Using Range, only 8*2 space is required.

OK I looked into it again, and it makes sense now. The mention of “maps” in the commit description is a bit confusing, but fine.
@wdullaer Is the API change on `register_allowed_memory()` in this PR an issue for you?

As for the tests:

> I don't think the lack of this test will affect the use of the JIT in no_std mode.

I don't think either. But then, the number of times I thought my code would work and it didn't! :slightly_smiling_face: No, if we support the JIT without `std`, we need to have proper tests to check that it does work, and to catch regressions. We have a nice suite of tests in rbpf, would be a shame not to use them to test the different features.

> It is a bit difficult to test the JIT in no_std environment, because executable memory allocation is usually kernel-specific. 

What does this mean exactly? Is it too hard to put in a wrapper and make it work with the different CI runners? What do you mean by “kernel-specific”, are there more cases than Linux/macOS and Windows?

Godones:
Maybe we have different understandings of no_std mode, but it doesn’t matter. Now we allow running jit tests in --no-default-features mode.

qmonnet (approved):
Looks all good to me. Thanks for this!

qmonnet:
I was about to merge but noticed the CI failures, which seem unrelated and should be addressed in https://github.com/qmonnet/rbpf/pull/126.

Would you mind rebasing so we can run the full tests with `--no-default-features`, please?

Godones:
> I was about to merge but noticed the CI failures, which seem unrelated and should be addressed in #126.
> 
> Would you mind rebasing so we can run the full tests with `--no-default-features`, please?

No problem.

Godones:
> > Maybe we have different understandings of no_std mode, but it doesn’t matter.
> 
> What do you mean? (Genuine question, I may be wrong and would like to understand better.)
> 
> > Now we allow running jit tests in --no-default-features mode.
> 
> Thanks a lot for this!

I think the `no_std` mode is generally suited for typical kernel scenarios, such as the Linux kernel or other embedded systems. In these cases, `no_std` means we can't use the standard library or rely on `libc` support. However, from a literal perspective, `no_std` simply means that we cannot use the standard library or any libraries that depend on it. So, we can still use `libc` to complete the test in user mode.

PR #127, opened 2025-06-30
jit: Enable JIT without "std" feature, but with global alloc

nathanjsweet:
Add a new "alloc" feature that indicates that the crate importing rbpf has an implementation of the GlobalAlloc trait. Memory protection is ignored in the "no_std" environment.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         toolchain: [stable, beta, nightly]
-        features: [--all-features]
+        features: [--features "std cranelift"]
         include:
           - toolchain: nightly
             features: --no-default-features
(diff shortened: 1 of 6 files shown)


Godones:
In this case, the rbpf library cannot ensure that the memory has executable permissions. Developers need to ensure that the memory allocated by GlobalAlloc has executable permissions, which seems unreasonable and dangerous.

qmonnet (changes requested):
Sorry for the delay - took me some time to look into this. Thank you also Linfeng for the feedback.

I agree with Nate, it sounds acceptable to leave the user responsible for checking memory permissions, given that the `alloc` feature is opt-in.

Maybe we could add a comment about this in the feature description in the README, though?

Code looks good otherwise, with just some typos in the README update.

Could you please add a full test or example to show how to use the feature? (It doesn't have to be a complex one, but something providing a basic implementation for `GlobalAlloc` and running a JITed program on that would help users (and maintainers :innocent:) understand how to use it.

qmonnet:
We also need to adjust the CI runs, where we build with `--all-features`. Is there a way to make `--all-features` do all-features-but-actually-not-mutually-exclusive-ones in Rust/Cargo?

PR #138, opened 2026-02-02
fix: cast i32 to u32 first for zero ext

hsqStephenZhang:
Should cast i32 value to u32 first for zero extension.

The fix is quite straight forward, the test cases are added. It will panic before the fix.


qmonnet (approved):
Thank you!

PR #139, opened 2026-02-02
fix: init reg[2] as mbuff.len()

hsqStephenZhang:


qmonnet:
Thanks for the context. I see that https://github.com/iovisor/ubpf/issues/57 says “The eBPF spec allows R1-R5 to be used for such purposes”. I believe it refers to [this draft](https://www.ietf.org/archive/id/draft-thaler-bpf-isa-00.html#section-2-5):

> Upon entering execution of an eBPF program, registers R1 - R5 initially can contain the input arguments for the program (similar to the argc/argv pair for a typical C program). The actual number of registers used, and their meaning, is defined by the program type; for example, a networking program might have an argument that includes network packet data and/or metadata.

But this convention never explicitly made it to [the final ISA specification](https://www.rfc-editor.org/rfc/rfc9669.html#name-introduction). So there's no real convention as far as I know.

This change probably makes more sense for uBPF, which doesn't work with `mbuff`s as far as I remember. I think I'll stick to the kernel behaviour in rbpf for now, unless there's a strong demand for use cases that `mbuff` cannot addressed.

hsqStephenZhang:
Got it. Thanks!

PR #140, opened 2026-02-02
fix: use is_reg for devisor in emit_muldivmod

hsqStephenZhang:
should not use `ins.imm != 0` to  decide if it's a register operation

see: https://github.com/iovisor/ubpf/blob/d8af62f3a5f720d74a1d84ce6b94b86fb7c7cb10/vm/ubpf_jit_x86_64.c#L1299

### DIFF
--- a/src/jit.rs
+++ b/src/jit.rs
@@ -433,7 +433,9 @@ impl JitCompiler {
         if dst != RDX {
             self.emit_push(mem, RDX);
         }
-        if imm != 0 {
+
+        // Load the divisor into RCX.
+        if !is_reg {
             self.emit_load_imm(mem, RCX, imm as i64);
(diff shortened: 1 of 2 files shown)


qmonnet (changes requested):
Thanks for this!

The change looks good. This code was initially derived from uBPF, but there was indeed an issue: https://github.com/iovisor/ubpf/issues/448 and Alan addressed it in https://github.com/iovisor/ubpf/pull/449.

Can you please sign-off your commit?

qmonnet (approved):
Thanks!

hsqStephenZhang:
Hi, would you mind publishing a new patch version?

qmonnet:
Sure, done

PR #145, opened 2026-05-06
feat: legacy XADD and correct JIT code buffer sizing

sakateka:
Hi, and thanks for a great project.

I hit a wall: my program uses atomic operations and wouldn’t run- it was hitting unimplemented. After digging through the code, adding support for the legacy XADD instructions looked straightforward, so I went ahead and implemented it.

I might come back later with full atomic support, but for now I’d like to get this merged.

Thanks!

### DIFF
--- a/src/verifier.rs
+++ b/src/verifier.rs
@@ -154,8 +154,23 @@ pub fn check(prog: &[u8]) -> Result<(), Error> {
             ebpf::ST_H_REG   => store = true,
             ebpf::ST_W_REG   => store = true,
             ebpf::ST_DW_REG  => store = true,
-            ebpf::ST_W_XADD  => { reject(format!("XADD instructions are not supported (insn #{insn_ptr:?})"))?; },
-            ebpf::ST_DW_XADD => { reject(format!("XADD instructions are not supported (insn #{insn_ptr:?})"))?; },
+            ebpf::ST_W_XADD  => {
+                // Same opcode as BPF atomic stores; legacy XADD uses imm == 0.
+                if insn.imm != 0 {
+                    reject(format!(
+                        "atomic operations other than legacy XADD are not supported (insn #{insn_ptr:?})"
+                    ))?;
+                }
+                store = true;
+            },
+            ebpf::ST_DW_XADD => {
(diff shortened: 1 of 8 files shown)


sakateka:
Hi, @qmonnet! Could you please take a look at this PR when you have a moment? Thanks!

qmonnet:
Yep sorry, it's been a busy few weeks, but this is still on my list. I'll try to take a deeper look at it soon.

src/verifier.rs
```diff
@@ -154,8 +154,8 @@ pub fn check(prog: &[u8]) -> Result<(), Error> {
             ebpf::ST_H_REG   => store = true,
             ebpf::ST_W_REG   => store = true,
             ebpf::ST_DW_REG  => store = true,
-            ebpf::ST_W_XADD  => { reject(format!("XADD instructions are not supported (insn #{insn_ptr:?})"))?; },
-            ebpf::ST_DW_XADD => { reject(format!("XADD instructions are not supported (insn #{insn_ptr:?})"))?; },
+            ebpf::ST_W_XADD  => store = true,
+            ebpf::ST_DW_XADD => store = true,
```
qmonnet: This makes the legacy XADD work, but I think this will also allow programs with newer atomic operations to pass the checks, although we don't support the newer atomic yet. We should probably do `store = true` only if `insn.imm == 0`, and `reject()` otherwise?

qmonnet (changes requested):
Thanks, and apologies again for the delay.

It looks good overall, but I've got two comments:

- Can you please adjust the verifier check to keep rejecting newer atomic operations (see comment below)?
- Could you please make the JIT allocation fix a separate commit? It would make it easier to look at it again in the future, if necessary.

There's also `test_jit_xadd_w_dw()` which currently doesn't pass with `--no-default-features` (`"Error: No custom executable memory set, call set_jit_exec_memory() to set one"`).

sakateka:
Thanks for the hints and explanations!
I split the changes into separate commits, fixed the failing test, and also tried to improve coverage.

qmonnet (approved):
Looks good now, much better with two commits. Thank you!

PR #147, opened 2026-05-29
feat: add RISC-V 64-bit JIT backend

CN-TangLin:
## Description
This PR adds a complete RISC-V 64-bit (rv64) backend for the JIT compiler. 

### Key Features
- Full support for RV64 instruction set (ALU, ALU64, jumps, memory operations).
- Two-pass compilation strategy (calculating instruction sizes, then emitting machine code).
- Proper handling of BPF calling conventions and stack frame alignment.
- Added zero-division protection for `BPF_DIV` and `BPF_MOD`.
- Unified 32-bit zero-extension logic for ALU32 operations.

### Testing
- Passed existing architecture-independent tests.
- Added 20 new `#[cfg(target_arch = "riscv64")]` specific unit tests for instruction generation validation.
- Successfully integrated and tested within StarryOS (an OS kernel).

Looking forward to your review!


qmonnet (changes requested):
Hi, and thanks!

I haven't been through the code in details yet, but please:

- Drop the formatting changes from your first commit, they don't seem relevant and they make it harder to review. If they come from rustfmt, I'm OK to take them, but as a separate commit.
- Motivate the change in the second commit. Why do you need the `jit` module to be public? (The reason should also appear in the commit description)

qmonnet:
Thanks, the split will make it easier to review. I haven't had time to go through the details of the backend yet.

One question: Did you notice that there is a (not-really-documented) `cranelift` feature, which brings a JIT-compiler for the local architecture, and already has, as far as I understand, a backend for riscv64? Would it be enough for your use case, or is there a specific advantage to use a separate backend? I won't be able to really maintain the riscv64 JIT if it breaks so if this can be addressed with the existing feature, it would be just as well.

